Issue #24, opened 2015-03-01
Implementation feedback


guybedford:
I've got a working implementation at https://github.com/ModuleLoader/es6-module-loader/blob/1.0/src/loader.js (PR https://github.com/ModuleLoader/es6-module-loader/pull/317).

Here are some notes of the minor corrections from the process:

* 4.1.5 1 Instantiation must be called with a loader argument.
* 4.1.5 1 _instance_ is both an argument and a local variable.
* 4.1.5 4 Resolve _entry_.[[Instantiate]] with _instance_. We're resolving the promise from outside of the scope of the promise which practically, means storing the instantiateResolve function on the entry. It can be easier to move this out to a `return entry.module` in 4.2.3 at the instantiate promise return and a Promise.resolve in the last line of 4.1.4.
* 4.1.5 6 b For each _dep_ in _instance_.[[ImportedModules]] is this a ReflectiveModuleRecord property? I couldn't see it defined anywhere.
* 4.2.1 6 use _payload_ instead of _v_ for consistency?
* 4.2.3 5 Should RequestTranslate directly, don't need RequestFetch as it is implicit in RequestTranslate
* 4.2.3 Let _p1_ be the result of promise-calling _hook_(_key_, _payload_, _source_)
  Surely we just need _source_ and not _payload_?
* 4.2.4 _entry_ intead of _entry0_
* 4.2.4 Resolve is not defined anywhere? loader.[[resolve]]?
* 4.2.4 Before running RequestInstantiate for dependencies, we need to do the check:
  ```
    _depEntry_ = ensureRegistered(loader, depKey, metadata);
    if _depEntry_.[[state]] is "ready" then use _depEntry_ for _dep_
  ```
  Otherwise requestInstantiate throws if the module is already defined
* 4.2.5 _module_.Evaluate is not defined?
  Resolve is not defined?
* 5.2.1 Let _status_ be _module_.Instantiate(). I'm wondering about using the module object as both the execute function and the module instance?
* 6.3.1 To confirm - if stage is fetch or translate, it never proceeds to next stage?
* 6.4.2 Not sure how loader.prototype.error is to be implemented.
* 6.5.2 Should probably set error property to null for installed entry.
* 6.5.3 Reflect.loader.uninstall(key) not Reflect.loader.uninstall(module)

I also implemented the metadata piping. Having metadata available as soon as the resolve hook runs would really be the ideal here, but does mean that the first resolution takes precedence, which I know may not be ideal for the spec process.

[...]

Issue #25, opened 2015-03-01
Dependency Loading Bug


guybedford:
Currently in 4.2.4 RequestLink it calls `requestInstantiate` on all dependencies. This in turn will fulfill when all the dependencies have been driven to the LINK state. But just because the dependencies are in the link state does not mean that their dependencies in turn are. Thus third-level dependencies can be skipped out.

Let me know if that describes the issue in enough detail.

dherman:
@guybedford Can you take a look at commit 3dd0c66c34aa20c7454283a53482280aa5a3bef5 and see if that looks right to you?

guybedford:
@dherman this looks great to me on first read. Will work on updating this implementation next week with some luck.

caridy:
@guybedford are we good here? if yes, let's close this issue.

guybedford:
@caridy thanks for following up. Yes this is still a bug unfortunately, see the last point in https://github.com/whatwg/loader/issues/40, which is a critical bug wrt to circular references:

```
4.2.4 Circular references still don't have a stop condition here. Some type of extra state is needed, perhaps an INSTANTIATE_ALL that fits in after instantiate, which then gets marked as LINK within instantiateAll before hitting the circular references.
```

caridy:
ok, we will track the circular references issue here: https://github.com/whatwg/loader/issues/40

PR #26, opened 2015-03-03
implementation feedback from issue #24 - first pass

caridy:
> 4.1.5 1 Instantiation must be called with a loader argument.

`CommitInstantiated` is called from `RequestInstantiate` and `ResolveInstantiate`, which they both have access to `loader`. It needs loader to call `Instantiation()`.

> 4.1.5 1 instance is both an argument and a local variable.

renamed to `_module_`

> 4.2.1 6 use payload instead of v for consistency?

_payload_

> 4.2.3 5 Should RequestTranslate directly, don't need RequestFetch as it is implicit in RequestTranslate

the call to `RequestFetch` cannot be skipped because we need _payload_ when calling `_loader_.[[Instantiate]]`

> 4.2.4 entry intead of entry0

_entry_

> 4.2.4 Before running RequestInstantiate for dependencies, we need to do the check:  _depEntry_ = ensureRegistered(loader, depKey, metadata); if _depEntry_.[[state]] is "ready" then use _depEntry_ for _dep_  Otherwise requestInstantiate throws if the module is already defined

algo refactored to cover this case

> 6.5.3 Reflect.loader.uninstall(key) not Reflect.loader.uninstall(module)

_key_

/cc @guybedford, @dherman

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -236,23 +236,23 @@ A <dfn>registry entry</dfn> is a record with the following fields:
 
 <pre emu-alg>
 1. If _entry_.[[Instantiate]] is *undefined*, then set _entry_.[[Instantiate]] to a new promise.
-1. Return CommitInstantiated(_entry_, _instance_, _source_).
+1. Return CommitInstantiated(_loader_, _entry_, _instance_, _source_).
 </pre>
 
-<h4 id="commit-instantiated">CommitInstantiated(entry, instance, source)</h4>
+<h4 id="commit-instantiated">CommitInstantiated(loader, entry, instance, source)</h4>
 
 <pre emu-alg>
-1. Let _instance_ be Instantiation(_instance_, _source_).
-1. ReturnIfAbrupt(_instance_).
-1. // TODO: edge case: what if _instance_ is a thenable function?
-1. Resolve _entry_.[[Instantiate]] with _instance_.
+1. Let _module_ be Instantiation(_loader_, _instance_, _source_).
+1. ReturnIfAbrupt(_module_).
+1. // TODO: edge case: what if _module_ is a thenable function?
+1. Resolve _entry_.[[Instantiate]] with _module_.
 1. Let _deps_ be a new empty List.
-1. If _instance_ is a Module Record, then:
-  1. Assert: _instance_ is a Normal Module Record.
-  1. For each _dep_ in _instance_.[[ImportedModules]], do:
+1. If _module_ is a Module Record, then:
+  1. Assert: _module_ is a Normal Module Record.
+  1. For each _dep_ in _module_.[[ImportedModules]], do:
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -306,14 +306,13 @@ A <dfn>registry entry</dfn> is a record with the following fields:
 1. If _entry_.[[State]] is "ready", return a new error promise.
 1. If _entry_.[[Instantiate]] is not *undefined*, return _entry_.[[Instantiate]].
 1. Let _hook_ be _loader_.[[Instantiate]].
-1. Let _p_ be the result of transforming RequestFetch(_loader_, _key_) with a fulfillment handler that, when called with argument _payload_, runs the following steps:
-  1. Return the result of transforming RequestTranslate(_loader_, _key_) with a fulfillment handler that, when called with argument _source_, runs the following steps:
-    1. // TODO: metadata
-    1. Let _p1_ be the result of promise-calling _hook_(_key_, _payload_, _source_).
```
guybedford: Looking at this again - I guess that the `_payload_` comes from fetch and the `_source_` from translate which would explain this signature. So it looks like the instantiate hook is called with arguments - `key`, `originalSource`, `translatedSource`? May be worth moving this out for further clarification for now.

index.bs
```diff
@@ -321,12 +320,17 @@ A <dfn>registry entry</dfn> is a record with the following fields:
...
+1. If _entry_.[[State]] is "ready", return a new promise fulfilled with _entry_.[[Module]].
 1. Return the result of transforming RequestInstantiate(_loader_, _key_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
   1. Let _depLoads_ be a new empty List.
   1. For each _pair_ in _entry_.[[Dependencies]], do:
     1. Let _p_ be the result of transforming Resolve(_loader_, _pair_.[[key]], _key_) with a fulfillment handler that, when called with value _depKey_, runs the following steps:
+      1. Let _depEntry_ be EnsureRegistered(_loader_, _depKey_).
+      1. If _depEntry_.[[State]] is "ready", then:
+        1. Let _dep_ be _depEntry_.[[Module]].
+        1. Set _pair_.[[value]] to _dep_.
+        1. Return _dep_.
```
guybedford: The above change would likely be tackled along with https://github.com/whatwg/loader/issues/25 so perhaps wait on this for now as well.

guybedford:
Thanks @caridy for the quick review!

index.bs
caridy: yeah, will rollback this block, I missed that detail.

index.bs
caridy: I prefer to have this one in, since this is a bug in the current algo, and then do the refactor once #25 is cleared.

PR #57, opened 2015-07-29
Fixes #1: switch to ecmarkup

### LINKED ISSUE #1: switch to ecmark{up,down}
Once the tools are ready for use we'll want to use ecmarkup and ecmarkdown to do the algorithm steps.

caridy:
* [x] adding ecmarkup step into the build.
 * [x] algos using `<emu-alg>`
 * [x] adding watching capabilities for convenience `npm run local-watch`
 * [x] build steps in readme
 * [x] Makefile updates

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,39 @@
 # Loader
 
-This repository provides the [JavaScript Loader Standard](http://whatwg.github.io/loader).
-This consolidates work on the ECMAScript module loading semantics with the integration
-points of Web browsers, as well as Node.js.
+This repository provides the [JavaScript Loader Standard](http://whatwg.github.io/loader). This consolidates work on the ECMAScript module loading semantics with the integration points of Web browsers, as well as Node.js.
+
+The main spec is available at http://whatwg.github.io/loader, generated from the `index.bs` file.
+
+## Building the spec
+
+Building the spec is a two-step process. First, the majority of the conversion work is done via [Bikeshed](https://github.com/tabatkins/bikeshed). Second, we run a custom portion of the [Ecmarkup](https://github.com/bterlson/ecmarkup) pipeline to convert the algorithms from [Ecmarkdown](https://github.com/domenic/ecmarkdown) syntax into HTML, and to automatically add cross-references. This second step requires [io.js](https://iojs.org/) to be installed.
+
+### Bikeshed
+
+To run Bikeshed locally, [install Bikeshed](https://github.com/tabatkins/bikeshed/blob/master/docs/install.md) and then run `bikeshed spec` in the working directory.
+
+Alternately, you can use the command
+
+```
+curl https://api.csswg.org/bikeshed/ -F file=@index.bs > index.tmp.html
+```
+
+to use Bikeshed's web interface without installing anything.
+
+### Ecmarkup
+
+To run the Ecmarkup step, be sure you've done `npm install` in the root directory, then run
+
+```
+npm run ecmarkupify index.tmp.html index.html
+```
+
+### Local Watch
+
+If you have Bikeshed installed locally, and have run `npm install`, you can try running
+
+```
+npm run local-watch
+```
+
+to start a watcher on `index.bs` that will update `index.html` as you edit.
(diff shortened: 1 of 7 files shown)


index.bs
```diff
@@ -16,11 +18,21 @@ Abstract: loading process and customizing loading behavior.
...
+  .note + .example, .note + .note { margin-top: 1em; }
+
+  emu-val { font-weight: bold; }
+  emu-alg > ol, emu-alg > ol ol ol ol { list-style-type: decimal; }
+  emu-alg > ol ol, emu-alg > ol ol ol ol ol { list-style-type: lower-alpha; }
+  emu-alg > ol ol ol, emu-alg > ol ol ol ol ol ol { list-style-type: lower-roman; }
+  emu-alg li { margin: 0; }
 </style>
+<script src="https://resources.whatwg.org/file-issue.js" async></script>
+<script src="https://resources.whatwg.org/commit-snapshot-shortcut-key.js" async></script>
```
caridy: this is not really needed as right now, but eventually we will probably want to have the snapshots, I can remove it from now I guess.

README.md
```diff
@@ -1,5 +1,39 @@
...
+Building the spec is a two-step process. First, the majority of the conversion work is done via [Bikeshed](https://github.com/tabatkins/bikeshed). Second, we run a custom portion of the [Ecmarkup](https://github.com/bterlson/ecmarkup) pipeline to convert the algorithms from [Ecmarkdown](https://github.com/domenic/ecmarkdown) syntax into HTML, and to automatically add cross-references. This second step requires [io.js](https://iojs.org/) to be installed.
+
+### Bikeshed
+
+To run Bikeshed locally, [install Bikeshed](https://github.com/tabatkins/bikeshed/blob/master/docs/install.md) and then run `bikeshed spec` in the working directory.
+
+Alternately, you can use the command
+
+```
+curl https://api.csswg.org/bikeshed/ -F md-Text-Macro="SNAPSHOT-LINK dummy" -F file=@index.bs > index.tmp.html
```
domenic: You probably don't need snapshot-related stuff; your makefile doesn't seem to have the facility for producing it. Here and in the .bs file.

index.bs
```diff
@@ -356,17 +368,17 @@ A <dfn>registry entry</dfn> is a record with the following fields:
   1. ReturnIfAbrupt(_status_).
   1. Assert: _entry_'s whole dependency graph is in "ready" state.
   1. Return _entry_.
-</pre>
+</emu-alg>
 
 <h4 id="request-ready">RequestReady(loader, key)</h4>
 
-<pre emu-alg>
+<emu-alg aoid="RequestReady">
```
domenic: I put the aoids on the h4s; seems nicer

index.bs
```diff
@@ -701,19 +712,19 @@ Reflective modules are always already instantiated.
...
-<pre emu-alg>
+<emu-alg aoid="CreateModuleMutator">
 1. // TODO: spec me
-</pre>
+</emu-alg>
 
 <h4 id="reflect-module">Reflect.Module(descriptors[, executor, evaluate])</h4>
 
-<pre emu-alg>
+<emu-alg aoid="Reflect.Module">
```
domenic: No aoids for public APIs

index.bs
caridy: yeah, I noticed that. It is definitely nicer since the link will jump to the header instead of the first line of the algo. I can do that.

index.bs
caridy: that was one of my open questions. Does the same applies to something like `Module.GetExportNames`?

domenic: I've never seen a precedent for "Reflective Module.GetExportNames" in the spec so I'm not sure. It seems like an abstract op to me so maybe aoid is OK.

caridy:
@domenic I think we are ready to merge this. We can add snapshots if needed.

As for fixing typos and other algo related stuff, we can do it in another PR that will require extensive review from @dherman.

domenic:
Cool, you got the styling issue you mentioned in person figured out?

caridy:
@domenic yes, as you predicted, there were few inline styles that were not ported into the loader doc from streams. We are good now.

PR #65, opened 2015-08-14
Enabling Subclassing of `Reflect.Loader` (major reorganization)

caridy:
_[compiled spec from this PR](https://rawgit.com/caridy/d189efc249d40f2242af/raw/490d9586e646b5fa48b7919d87f68e8e9204c941/loader-spec.html)_

Changes:

* [x] Well-lnown Symbols
* [x] Loader Objects (loader base class definition)
* [x] Registry Objects (internal class definition for loader base class' [[Registry]] internal slot)
* [x] Module Objects (base class definition for Modules, former Reflective Module)
* [x] Browser Loader Objects (loader class for default loader instances implemented by browsers)
* [x] `System.loader` is an instance of Browser Loader
* [x] SitePackage was deferred (removed from the spec for now)
* [x] Use symbols to access hooks
* [x] Iterate over Registry (e.g.: `for (let entry of System.loader.registry) {}`)
* [x] Avoid leaking internal temporary slots from registry entries. (more work to do here, for now, we expose `state` and `statePromise` instead of exposing all available internal promises for every past state)
* [x] Reflective Module -> Module Objects (Reflect.Module)
* [x] Backpointer from registry to the loader instance it belongs to.
* [x] Reflect.Loader.prototype.load() is now name and referrer driven.
* [x] annexes

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -174,7 +174,7 @@ The Reflect.Loader constructor is designed to be subclassable. It may be used as
 
 <h4 id="new-reflect-loader">Reflect.Loader()</h4>
 
-When Registry is called with no arguments, the following steps are taken:
+When Reflect.Loader is called with no arguments, the following steps are taken:
 
 <emu-alg>
 1. If NewTarget is *undefined*, then throw a *TypeError* exception.
@@ -837,7 +837,7 @@ The Module constructor is the initial value of the Module property of the the Re
 
 The Reflect.Module constructor is designed to be subclassable. It may be used as the value in an extends clause of a class definition. Subclass constructors that intend to inherit the specified Module behaviour must include a super call to the Reflect.Module constructor to create and initialize the subclass instance with the internal state necessary to support the Reflect.Module.prototype built-in methods.
 
-<h4 id="new-reflect-module">Reflect.Module(descriptors[, executor, evaluate])</h4>
+<h4 id="new-reflect-module">Reflect.Module(descriptors[, executor[, evaluate]])</h4>
 
 When Reflect.Module is called with arguments <i>descriptors</i>, <i>executor</i>, and <i>evaluate</i>, the following steps are taken:
 
--- a/index.bs
+++ b/index.bs
@@ -249,8 +249,74 @@ The following steps are taken:
 <emu-alg>
 1. Let _loader_ be *this* value.
 1. If Type(_loader_) is not Object, throw a *TypeError* exception.
-1. If _loader_ does not have a [[Registry]] internal slot throw a *TypeError* exception.
 1. Return Resolve(_loader_, _name_, _referrer_).
+</emu-alg>
+
+<h4 id="reflect-loader-load">Reflect.Loader.prototype.load(key[, stage])</h4>
+
+The following steps are taken:
+
+<emu-alg>
+1. Let _loader_ be *this* value.
+1. If Type(_loader_) is not Object, throw a *TypeError* exception.
+1. If _stage_ is *undefined* then let _stage_ be "ready".
+1. If _stage_ is "fetch", then:
+  1. Return RequestFetch(_loader_, _key_).
+1. If _stage_ is "translate", then:
+  1. Return RequestTranslate(_loader_, _key_).
+1. If _stage_ is "instantiate", then:
+  1. Return the result of transforming RequestInstantiateAll(_loader_, _key_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
+    1. If _entry_.[[Module]] is a Function object, return _entry_.[[Module]].
+    1. Return *undefined*.
+1. If _stage_ is "link", then:
+  1. Return the result of transforming RequestLink(_loader_, _key_) with a fulfillment handler that returns *undefined*.
+1. If _stage_ is "ready", then:
+  1. Return the result of transforming RequestReady(_loader_, _key_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
+    1. Return GetModuleNamespace(_entry_.[[Module]]).
+1. Throw a new *TypeError*.
+</emu-alg>
+
+<h4 id="reflect-loader-provide">Reflect.Loader.prototype.provide(key, stage, value)</h4>
+
+The following steps are taken:
+
+<emu-alg>
+1. Let _loader_ be *this* value.
+1. If Type(_loader_) is not Object, throw a *TypeError* exception.
+1. Let _entry_ be EnsureRegistered(_loader_, _key_).
+1. If _stage_ is "fetch", then:
+  1. Let _stateValue_ be GetStateValue(_entry_.[[State]]).
+  1. Let _fetchStateValue_ be GetStateValue("fetch").
+  1. If _stateValue_ is greater than _fetchStateValue_, throw a new *TypeError*.
+  1. Call FulfillFetch(_loader_, _entry_, _value_).
+  1. Return *undefined*.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -66,6 +66,88 @@ The primary goal is to make as much of this process as possible consistent betwe
 
 <h2 id="conventions">Conventions</h2>
 
+<h3 id="well-known-intrinsic-objects">Well-Known Intrinsic Objects</h3>
+
+Well-known intrinsics are built-in objects that are explicitly referenced by the algorithms of this specification and which usually have Realm specific identities. Unless otherwise specified each intrinsic object actually corresponds to a set of similar objects, one per Realm.
+
+Within this specification a reference such as %name% means the intrinsic object, associated with the current Realm, corresponding to the name. Determination of the current Realm and its intrinsics is described in ES2015, 8.3. The well-known intrinsics are listed in Table X.
```
dherman: italicize *name*, as in:

"a reference such as %*name*% means the intrinsic object, associated with the current Realm, corresponding to *name*."

index.bs
```diff
@@ -66,6 +66,88 @@ The primary goal is to make as much of this process as possible consistent betwe
...
+
+<table>
```
dherman: The conventions section is only establishing spec conventions used in the rest of the spec, but it doesn't contain actual spec contents. So I wouldn't put this table here.

Since this table really cuts across different categories (especially across things that are universal vs browser-specific), I would put it in an appendix at the end, with a cell that refers to the section of the spec that introduces that row's intrinsic.

index.bs
```diff
@@ -66,6 +66,88 @@ The primary goal is to make as much of this process as possible consistent betwe
...
+  <thead>
+    <tr>
+      <th>Intrinsic Name</th>
+      <th>Global Name</th>
+      <th>Description (<em>non-normative</em>)</th>
+    </tr>
+  </thead>
+  <tr>
```
dherman: I don't believe we need a %Loader% or %LoaderPrototype% intrinsic. `Reflect.Loader` and `Reflect.Loader.prototype` should not have any particular powers or connection to the realm, and nothing in the semantics needs to refer to them; they're just regular library functions like, say, `parseInt`.

index.bs
```diff
@@ -66,6 +66,88 @@ The primary goal is to make as much of this process as possible consistent betwe
...
+    <td>%BrowserLoader%</td>
+    <td></td>
+    <td>The browser loader constructor (<a href="#browser-loader">...</a>)</td>
+  </tr>
+  <tr>
+    <td>%BrowserLoaderPrototype%</td>
+    <td></td>
+    <td>The initial value of the *prototype* data property of %BrowserLoader%</td>
+  </tr>
+  <tr>
```
dherman: Given that site-packages are half-baked right now, and slated for the latest stage of the [roadmap](https://github.com/whatwg/loader/blob/master/roadmap.md), I think we should just eliminate the feature completely from the spec for now.

caridy: ok, will do.

index.bs
caridy: Got it, I will update the PR. As today, they are only used in BrowserLoader Objects section, but as you mentioned, I can just use `Reflect.Loader` and `Reflect.Loader.prototype` directly.

index.bs
caridy: ok

index.bs
```diff
@@ -899,13 +1030,11 @@ The browser loader has an extra internal slot:
   <li>top-level packages consult \[[Site]]: <code>"jquery"</code>, <code>"ember/data"</code>
```
dherman: delete this TODO since we're deferring site-packages

index.bs
```diff
@@ -736,28 +881,15 @@ Reflective modules are always already instantiated.
...
-  <li>re-export (immutable): <code>{ module: m, import: "foo" }</code>
-</ul>
-
-<emu-alg>
-1. // TODO: spec me
-</emu-alg>
+The Module constructor is the initial value of the Module property of the the Reflect object. When called as a constructor it creates and initializes a new Module object. Reflect.Module is not intended to be called as a function and will throw an exception when called in that manner.
 
-<h4 id="create-module-mutator" aoid="CreateModuleMutator">CreateModuleMutator(module)</h4>
+The Reflect.Module constructor is designed to be subclassable. It may be used as the value in an extends clause of a class definition. Subclass constructors that intend to inherit the specified Module behaviour must include a super call to the Reflect.Module constructor to create and initialize the subclass instance with the internal state necessary to support the Reflect.Module.prototype built-in methods.
```
dherman: change the end of this to "with the internal state necessary to integrated with loaders."

index.bs
```diff
@@ -798,97 +930,103 @@ Reflective modules are always already instantiated.
...
+This property has the attributes { \[[Writable]]: false, \[[Enumerable]]: false, \[[Configurable]]: false }.
 
-The browser loader contains extra properties for storing <dfn>site packages</dfn>, an application-global set of globally available packages. These map in an internal table to unique URLs that in turn serve as keys in the module registry.
+<h3 id="module-prototype-object">Properties of the Reflect.Module Prototype Object</h3>
 
-<div class="note">
-<p>
-The site package system serves as a simple coordination mechanism for modest-sized applications, but it does not provide all functionality required of a full-fledged package management system. It is expected that development ecosystems will build around package management tools that deal with requirements outside the scope of this specification, such as version management and allowing multiple versions of a library to coexist with the same name.
-</p>
+<h4 id="Reflect.Module.prototype.constructor">Reflect.Module.prototype.constructor</h4>
```
dherman: remove this -- there should not be a `constructor` property on the prototype (it can't have any properties).

index.bs
```diff
@@ -798,97 +930,103 @@ Reflective modules are always already instantiated.
...
-<p>
-Tools that preprocess JavaScript source code may choose to use or ignore the site package table. For example, a package manager may choose to preprocess two separate import statements requiring <code>"jquery"</code> to <code>"jquery/1.9"</code> and <code>"jquery/2.1.1"</code> respectively, based on configuration files informing the tool of version requirements. The tool would then store both versions of jQuery in the site package table using the longer names. Alternatively, the tool may choose to preprocess the imports directly as URLs and bypass the site package system altogether.
-</p>
-</div>
+The initial value of Reflect.Module.prototype.constructor is Reflect.Module.
 
-The browser loader has an extra internal slot:
+<h3 id="reflect-module-internal-slots">Properties of Module Instances</h3>
+
+Reflect.Module instances are ordinary objects that inherit properties from the Reflect.Module.prototype.
```
dherman: This isn't correct. Reflect.Module instances are "module namespace exotic objects" as defined in ES6. However, we should fix a slight design flaw in ES6, which is that in in 9.4.6.1, a module namespace exotic object always reports its prototype as null. We'll need to add a section in this spec describing the fix. We can do this in a separate PR.

index.bs
```diff
@@ -798,97 +930,103 @@ Reflective modules are always already instantiated.
...
+
+Reflect.Module instances are initially created with the internal slots described in the following table:
```
dherman: This stuff shouldn't get moved out of the section on Reflective Module Records. The \[[LocalExports]], \[[IndirectExports]], and \[[Evaluate]] internal properties belong on a Reflective Module Record, which is encapsulated internally by the module namespace exotic object returned by the `Reflect.Module` constructor.

dherman:
So much amazing work. Thanks @caridy, this is such great progress. <3

guybedford:
:+1:

PR #66, opened 2015-08-15
Result promise of RequestInstantiateAll will be fulfilled with the re…

Constellation:
Here, we assume that the result promise of the RequestInstantiateAll will be fulfilled
with the module record instance. But actually it will be fulfilled with the registry entry.
Since the result of the RequestInstantiate is also fulfilled with the registry entry,
we align the RequestInstantiateAll to the RequestInstantiate.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -370,7 +370,8 @@ A <dfn>registry entry</dfn> is a record with the following fields:
         1. Let _dep_ be _depEntry_.[[Module]].
         1. Set _pair_.[[value]] to _dep_.
         1. Return _dep_.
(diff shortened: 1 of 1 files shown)


Constellation:
Or is it better fulfilling `[[instantiate]]` and the result of `RequestInstantiateAll` with the module `instance` value?

Constellation:
In the meantime, since `RequestLink` also assumes that the result of `RequestInstantiateAll` will be fulfilled with the `entry`, I aligned to that.

caridy:
In general, we have a problem with fulfilling with registry entry, `[[Module]]` or `[[Namespace]]`, we need to revisit all of them to make sure we are doing the right thing. I will be taking care of it on PR #65.

caridy:
@Constellation html version was updated to reflect this fix.

Constellation:
Thank you!

PR #67, opened 2015-08-15
Separate "force fulfill" operation from the CommitInstantiated

Constellation:
While the RequestInstantiate fulfills `entry.[[Instantiate]]` promise
with the registry entry (as the returned value of the promise),
CommitInstantiated step 4 will fulfill it with the module instance proactively.

In this patch, we separate the "force fulfill" operation from the CommitInstantiated.
Move this operation to the FulFillInstantiate. And here, we will fulfill the `entry.[[Instantiate]]`
promise with the registry entry because the hander side requires it.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -271,7 +271,8 @@ A <dfn>registry entry</dfn> is a record with the following fields:
 
 <emu-alg>
 1. If _entry_.[[Instantiate]] is *undefined*, then set _entry_.[[Instantiate]] to a new promise.
-1. Return CommitInstantiated(_loader_, _entry_, _optionalInstance_, _source_).
+1. CommitInstantiated(_loader_, _entry_, _optionalInstance_, _source_).
+1. Fulfill _entry_.[[Instantiate]] with _entry_.
 </emu-alg>
 
 <h4 id="commit-instantiated" aoid="CommitInstantiated">CommitInstantiated(loader, entry, optionalInstance, source)</h4>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -280,7 +281,6 @@ A <dfn>registry entry</dfn> is a record with the following fields:
 1. Let _instance_ be Instantiation(_loader_, _optionalInstance_, _source_).
 1. ReturnIfAbrupt(_instance_).
 1. // TODO: edge case: what if _instance_ is a thenable function?
-1. Fulfill _entry_.[[Instantiate]] with _instance_.
```
caridy: `CommitInstantiated()` is called from:

* `RequestInstantiate()`, in which case the promise in `[[Instantiate]]` will be resolved implicitely, and there is not need to fulfilled manually.
* `FulfillInstantiate()`, in which case the promise should be fulfillled manually with the provided value.

This change seems to be correct.

Constellation:
Thanks for your reviews! I'll update the PR to reflect your comments.

dherman:
Recent overhauls to the spec have obsoleted this PR -- in particular, commit 261543c66c235b2a4b3a9e5076b6dd0386a94ec3 eliminates the explicit fulfillment of the \[[Instantiate]] promise, which is now simply piped through the pipeline thanks to Caridy's work in commit 2eff7f162b1d34ed1a2220c07e91d5b1c2513038.

PR #91, opened 2015-10-09
related to issue #88: refactor to use a list of stages instead of internal slots for each state.

caridy:
Changes:

* `state` is now `stage`in the pipeline, which is more clear.
* stages are stored in an ordered list, and each stage entry is removed as soon as the entry reaches a new stage.
* renaming `CommitInstantiate` to `ExtractDependencies`.

Rendered HTML:

* https://rawgit.com/caridy/148a3ffbb7caafc20da2/raw/eea60a38aa6ef84ed3384017eb82015af83d5b44/pr-91-loader.html

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -17,7 +17,7 @@ Abstract: JavaScript host environment. It also provides APIs for intercepting th
 Abstract: loading process and customizing loading behavior.
 Logo: https://resources.whatwg.org/logo-javascript.svg
 !Version History: <a href="https://github.com/whatwg/loader/commits">https://github.com/whatwg/loader/commits</a>
-!Participate: <a href="https://github.com/whatwg/loader/issues/new">File an issue</a> (<a href="https://github.com/whatwg/loader/issues?stage=open">open issues</a>)
+!Participate: <a href="https://github.com/whatwg/loader/issues/new">File an issue</a> (<a href="https://github.com/whatwg/loader/issues?state=open">open issues</a>)
 
 Opaque Elements: emu-alg
 </pre>
--- a/index.bs
+++ b/index.bs
@@ -19,7 +19,7 @@ Logo: https://resources.whatwg.org/logo-javascript.svg
 !Version History: <a href="https://github.com/whatwg/loader/commits">https://github.com/whatwg/loader/commits</a>
 !Participate: <a href="https://github.com/whatwg/loader/issues/new">File an issue</a> (<a href="https://github.com/whatwg/loader/issues?state=open">open issues</a>)
 
-Opaque Elements: emu-alg
+Opaque Elements: emu-alg, emu-note
 </pre>
 
 <style>
@@ -30,6 +30,8 @@ Opaque Elements: emu-alg
   emu-alg > ol ol, emu-alg > ol ol ol ol ol { list-style-type: lower-alpha; }
   emu-alg > ol ol ol, emu-alg > ol ol ol ol ol ol { list-style-type: lower-roman; }
   emu-alg li { margin: 0; }
+  emu-note { display: block; margin: 1em 0 1em 6em; color: #666; }
+  emu-note::before { content: "Note"; text-transform: uppercase; margin-left: -6em; display: block; float: left; }
 </style>
 <script src="https://resources.whatwg.org/file-issue.js" async></script>
 
@@ -317,6 +319,14 @@ The abstract operation GetStage with arguments entry and stage performs the foll
     1. Remove first element from _pipeline_.
 </emu-alg>
 
+<emu-note>
+  The internal slot [[Pipeline]] of an entry can never be empty.
+</emu-note>
+
+<emu-note>
+  Alternative, this algo can be implemented using functional programming techinques or a reverse cycle to avoid walking the entries twice.
+</emu-note>
+
 An <dfn id="stage-entry">stage</dfn> is a record with the following fields:
 
 <table>
@@ -354,7 +364,7 @@ The abstract operation GetRegistryEntry with arguments registry and key performs
 1. Let _stage_ be _currentStage_.[[Stage]].
 1. Let _result_ be CreateObject().
 1. Call SimpleDefine(_result_, "stage", _stage_).
-1. Call SimpleDefine(_result_, "stagePromise", _result_).
+1. Call SimpleDefine(_result_, "result", _result_).
 1. If _stage_ is "ready" then let _module_ be _entry_.[[Module]].
 1. Else let _module_ be *undefined*.
 1. Call SimpleDefine(_result_, "module", _module_).
@@ -589,7 +599,7 @@ Registry instances are initially created with the internal slots described in th
 1. If _pair_ exists, then:
   1. Let _entry_ be _pair_.[[value]].
 1. Else:
-  1. Let _pipeline_ to be a new List.
+  1. Let _pipeline_ be a new List.
   1. Add new stage entry record { [[Stage]]: "fetch", [[Result]]: undefined } as a new element of the list _pipeline_.
   1. Add new stage entry record { [[Stage]]: "translate", [[Result]]: undefined } as a new element of the list _pipeline_.
   1. Add new stage entry record { [[Stage]]: "instantiate", [[Result]]: undefined } as a new element of the list _pipeline_.
--- a/index.bs
+++ b/index.bs
@@ -349,11 +349,13 @@ The abstract operation GetRegistryEntry with arguments registry and key performs
 1. Let _pair_ be the entry in _entries_ such that _pair_.[[key]] is equal to _key_.
 1. If _pair_ does not exist, then return *null*.
 1. Let _entry_ be _pair_.[[value]].
-1. Let _result_ be CreateObject().
 1. Let _currentStage_ be GetCurrentStage(_entry_).
-1. Call SimpleDefine(_result_, "stage", _currentStage_.[[Stage]]).
-1. Call SimpleDefine(_result_, "stagePromise", _currentStage_.[[Result]]).
-1. If _currentStage_.[[Stage]] is "ready" then let _module_ be _entry_.[[Module]].
+1. Let _result_ be the result of transforming _currentStage_.[[Result]] with a new pass-through promise.
+1. Let _stage_ be _currentStage_.[[Stage]].
+1. Let _result_ be CreateObject().
+1. Call SimpleDefine(_result_, "stage", _stage_).
+1. Call SimpleDefine(_result_, "stagePromise", _result_).
+1. If _stage_ is "ready" then let _module_ be _entry_.[[Module]].
 1. Else let _module_ be *undefined*.
 1. Call SimpleDefine(_result_, "module", _module_).
 1. If _entry_.[[Error]] is *nothing*, then:
--- a/index.bs
+++ b/index.bs
@@ -17,7 +17,7 @@ Abstract: JavaScript host environment. It also provides APIs for intercepting th
 Abstract: loading process and customizing loading behavior.
 Logo: https://resources.whatwg.org/logo-javascript.svg
 !Version History: <a href="https://github.com/whatwg/loader/commits">https://github.com/whatwg/loader/commits</a>
-!Participate: <a href="https://github.com/whatwg/loader/issues/new">File an issue</a> (<a href="https://github.com/whatwg/loader/issues?state=open">open issues</a>)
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -153,32 +153,13 @@ mean the same thing as:
 1. Return the result of calling OrdinaryDefineOwnProperty(_obj_, _name_, _desc_).
 </emu-alg>
 
-<h4 id="get-state-value" aoid="GetStateValue">GetStateValue(state)</h4>
```
caridy: stages are now in an ordered list, not need to track down the numeric values.

index.bs
```diff
@@ -303,6 +284,61 @@ Loader instances are initially created with the internal slots described in the
 
 <h3 id="registry-abstract-operations">Abstract Operations for Registry Objects</h3>
 
+<h4 id="registry-GetCurrentStage" aoid="GetCurrentStage">GetCurrentStage(entry)</h4>
```
caridy: a couple of abstract operations to interact with the pipeline, which is the ordered list of stages.

index.bs
```diff
@@ -303,6 +284,61 @@ Loader instances are initially created with the internal slots described in the
...
+      <th>Description (<em>non-normative</em>)</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>\[[Stage]]</td>
+    <td><code>"fetch"</code>, <code>"translate"</code>, <code>"instantiate"</code>, <code>"link"</code>, <code>"ready"</code></td>
+    <td>A constant value to indicating which phase the entry is at.</td>
+  </tr>
+  <tr>
+    <td>\[[Result]]</td>
```
caridy: Question: I think we should use `[[Value]]` instead of `[[Result]]`, it is not really a result, but a thenable value.

index.bs
```diff
@@ -314,18 +350,10 @@ The abstract operation GetRegistryEntry with arguments registry and key performs
...
-1. Else If _entry_.[[State]] is "translate" and _entry_.[[Translate]] is not *undefined*, then
-  1. Set _statePromise_ to the result of transforming _entry_.[[Translate]] with a new pass-through promise.
-1. Else If _entry_.[[State]] is "instantiate" and _entry_.[[Instantiate]] is not *undefined*, then
-  1. Set _statePromise_ to the result of transforming _entry_.[[Instantiate]] with a fulfillment handler that, when called with argument _entry_, runs the following steps:
-    1. If _entry_.[[Module]] is a Function object, then return _entry_.[[Module]].
-    1. Return *undefined*.
-1. Call SimpleDefine(_result_, "statePromise", _statePromise_).
-1. If _entry_.[[State]] is "ready" then let _module_ be _entry_.[[Module]].
+1. Let _currentStage_ be GetCurrentStage(_entry_).
+1. Call SimpleDefine(_result_, "stage", _currentStage_.[[Stage]]).
```
caridy: Question: right now we are exposing `stage` and `stagePromise` as part of the proxy to the registry entry. I think we can make this `stage` and `value`, which is just the thenable value for the entry ready stage, or the current stage.

index.bs
```diff
@@ -313,19 +349,13 @@ The abstract operation GetRegistryEntry with arguments registry and key performs
...
-  1. Set _statePromise_ to the result of transforming _entry_.[[Fetch]] with a new pass-through promise.
-1. Else If _entry_.[[State]] is "translate" and _entry_.[[Translate]] is not *undefined*, then
-  1. Set _statePromise_ to the result of transforming _entry_.[[Translate]] with a new pass-through promise.
-1. Else If _entry_.[[State]] is "instantiate" and _entry_.[[Instantiate]] is not *undefined*, then
-  1. Set _statePromise_ to the result of transforming _entry_.[[Instantiate]] with a fulfillment handler that, when called with argument _entry_, runs the following steps:
-    1. If _entry_.[[Module]] is a Function object, then return _entry_.[[Module]].
-    1. Return *undefined*.
-1. Call SimpleDefine(_result_, "statePromise", _statePromise_).
-1. If _entry_.[[State]] is "ready" then let _module_ be _entry_.[[Module]].
+1. Call SimpleDefine(_result_, "stage", _stage_).
```
caridy: Question: right now we are exposing `stage` and `stagePromise` as part of the shim for the registry entry. I think we can make this `stage` and `value`, which is just the thenable value for the entry's "ready" stage, or the current stage.

index.bs
```diff
@@ -353,9 +383,9 @@ A <dfn id="registry-entry">registry entry</dfn> is a record with the following f
     <td>The resolved module key.</td>
   </tr>
   <tr>
-    <td>\[[State]]</td>
-    <td><code>"fetch"</code>, <code>"translate"</code>, <code>"instantiate"</code>, <code>"link"</code>, <code>"ready"</code></td>
-    <td>A constant value to indicating which phase the entry is at.</td>
+    <td>\[[Pipeline]]</td>
```
caridy: Using the list of stages instead of individual internal slots.

index.bs
```diff
@@ -465,7 +480,10 @@ The following steps are taken:
 1. Let _entries_ be _registry_.[[RegistryData]].
 1. Let _pair_ be the entry in _entries_ such that _pair_.[[key]] is equal to _key_.
 1. If _pair_ exists, then throw a new *TypeError*.
-1. Let _entry_ be a new registry entry record { [[Key]]: _key_, [[State]]: "ready", [[Metadata]]: *undefined*, [[Fetch]]: *undefined*, [[Translate]]: *undefined*, [[Instantiate]]: *undefined*, [[Dependencies]]: *undefined*, [[Module]]: _module_ }.
+1. Let _result_ be a new promise resolved with _module_.
```
caridy: directly creating the new record in "ready" stage, not need to add all other stages.

index.bs
```diff
@@ -479,9 +497,9 @@ The following steps are taken:
 1. Let _entries_ be _registry_.[[RegistryData]].
 1. Let _pair_ be the entry in _entries_ such that _pair_.[[key]] is equal to _key_.
 1. If _pair_ does not exist, then throw a new *TypeError*.
-1. Let _stateValue_ be GetStateValue(_pair_.[[value]].[[State]]).
-1. Let _linkStateValue_ be GetStateValue("link").
-1. If _stateValue_ is less than _linkStateValue_, then throw a new *TypeError*.
```
caridy: Question: I'm not sure about this condition. Why can we uninstall entries that are in `fetch`, `translate` or `instantiate`?

index.bs
```diff
@@ -303,6 +284,61 @@ Loader instances are initially created with the internal slots described in the
...
+<emu-alg>
+1. If Type(_entry_) is not Object, throw a *TypeError* exception.
+1. Let _stages_ be _entry_.[[Pipeline]].
+1. For each element _entry_ of _stages_, do
+  1. If _entry_.[[Stage]] is equal to _stage_, then
+    1. Return _entry_.
+1. Return *undefined*.
+</emu-alg>
+
+<h4 id="registry-UpgradeToStage" aoid="UpgradeToStage">UpgradeToStage(entry, stage)</h4>
```
caridy: this is the important key in this PR. With `UpgradeToStage()` we can simply drop old stages from the pipeline when reaching new stages, we don't need to worry about removing the last one, the "ready" stage will always we preserved, all other stages will be collected.

index.bs
```diff
@@ -543,7 +540,15 @@ The following steps are taken:
 The following steps are taken:
 
 <emu-alg>
-1. // TODO
+1. Let _registry_ be *this* value.
```
caridy: very similar to `provide()`, just rejecting...

index.bs
```diff
@@ -584,7 +589,13 @@ Registry instances are initially created with the internal slots described in th
 1. If _pair_ exists, then:
   1. Let _entry_ be _pair_.[[value]].
 1. Else:
-  1. Let _entry_ be a new registry entry record { [[Key]]: _key_, [[State]]: "fetch", [[Metadata]]: *undefined*, [[Fetch]]: *undefined*, [[Translate]]: *undefined*, [[Instantiate]]: *undefined*, [[Dependencies]]: *undefined*, [[Module]]: *undefined*, [[Error]]: *nothing* }.
+  1. Let _pipeline_ to be a new List.
+  1. Add new stage entry record { [[Stage]]: "fetch", [[Result]]: undefined } as a new element of the list _pipeline_.
```
caridy: I'm not very happy with this construction, open to a better way to define default entries for a list. we can't not formalize it in the structure because sometimes we create it with a single entry (e.g.: installing into "ready" stage).

index.bs
```diff
@@ -596,30 +607,7 @@ Registry instances are initially created with the internal slots described in th
 1. Return the result of promise-calling _hook_(_name_, _referrer_).
 </emu-alg>
 
-<h4 id="fulfill-fetch" aoid="FulfillFetch">FulfillFetch(loader, entry, payload)</h4>
```
caridy: finally we are getting rid of these methods, there is no more book keeping, everything happens naturally thru the pipeline.

index.bs
```diff
@@ -596,30 +607,7 @@ Registry instances are initially created with the internal slots described in th
...
-
-<h4 id="fulfill-instantiate" aoid="FulfillInstantiate">FulfillInstantiate(loader, entry, optionalInstance, source)</h4>
-
-<emu-alg>
-1. If _entry_.[[Instantiate]] is *undefined*, then set _entry_.[[Instantiate]] to a new promise.
-1. Return CommitInstantiated(_loader_, _entry_, _optionalInstance_, _source_).
-</emu-alg>
-
-<h4 id="commit-instantiated" aoid="CommitInstantiated">CommitInstantiated(loader, entry, optionalInstance, source)</h4>
+<h4 id="extract-dependencies" aoid="ExtractDependencies">ExtractDependencies(loader, entry, optionalInstance, source)</h4>
```
caridy: Renaming `CommitInstantiate` to `ExtractDependencies`, yay!

index.bs
```diff
@@ -303,6 +284,61 @@ Loader instances are initially created with the internal slots described in the
...
 <h3 id="registry-abstract-operations">Abstract Operations for Registry Objects</h3>
 
+<h4 id="registry-GetCurrentStage" aoid="GetCurrentStage">GetCurrentStage(entry)</h4>
+
+The abstract operation GetCurrentStage with argument entry performs the following steps:
+
+<emu-alg>
+1. If Type(_entry_) is not Object, throw a *TypeError* exception.
+1. Let _stages_ be _entry_.[[Pipeline]].
+1. Return the first element of _stages_.
```
dherman: add an assert: _stages_ is non-empty.

caridy: it can't be empty. it is created by EnsureRegistered() or .install(), either way you can't never make it empty, you can only upgrade to an existent stage.

index.bs
dherman: Let's stick with \[\[Result\]\]. The fact that it's a promise is an implementation detail; the name describes the thing that the promise computes.

index.bs
dherman: `stage` and `result`

index.bs
dherman: Yeah the registry state manipulation is a little funky. Let's deal with that after this PR and just leave it as-is for now.

index.bs
```diff
@@ -584,7 +589,13 @@ Registry instances are initially created with the internal slots described in th
 1. If _pair_ exists, then:
   1. Let _entry_ be _pair_.[[value]].
 1. Else:
-  1. Let _entry_ be a new registry entry record { [[Key]]: _key_, [[State]]: "fetch", [[Metadata]]: *undefined*, [[Fetch]]: *undefined*, [[Translate]]: *undefined*, [[Instantiate]]: *undefined*, [[Dependencies]]: *undefined*, [[Module]]: *undefined*, [[Error]]: *nothing* }.
+  1. Let _pipeline_ to be a new List.
```
dherman: Let _pipeline_ be

index.bs
dherman: It's not beautiful but it seems OK to me.

dherman:
Looks great. I have a couple tiny nits in the comments. Otherwise r+

index.bs
caridy: ok

index.bs
caridy: changed.

index.bs
caridy: ok

index.bs
caridy: added a note about it.

PR #93, opened 2015-10-15
provide() and error() are only allow for fetch, translate and instantiate stages

caridy:
https://github.com/whatwg/loader/pull/91 introduce a regression in the `provide()` and `error()` algos, this PR fixes that.


dherman:
Looks perfect, r+

Issue #68, opened 2015-08-15
Introducing the new state, "ResolveDependencies"


Constellation:
I think that the current draft spec incurs infinite recursion when there is the circular dependencies.
The situation is the following,

There are several modules,

__A__: This is the entry point.
```js
import "B"
```

__B__
```js
import "C"
```

__C__
```js
import "B"
```

1. At that time, before linking the modules, we call `RequestInstantiateAll(A)`.
2. In `RequestInstantiateAll(A)`, step 1-b-i-1, we extract the B dependency.
3. So, step 1-b-i-3, we perform `RequestInstantiateAll(B)`.
4. Later, it will extract the C dependency. So, step 1-b-i-3 in `RequestInstantiateAll(B)`, we call `RequestInstantiateAll(C)`.
5. Since C has the B dependency, in `RequestInstantiateAll(C)` step 1-b-i-3, we again call `RequestInstantiateAll(B)`.
6. Here, now, since once `RequestInstantiate(B)` is called inside `RequestInstantiateAll(B)`, it immediately returns the promise `B.[[Instantiate]]`. But, Now, __we stored the status of instantiation in [[Instantiate]], but we didn't store the status of collecting the dependencies__.
7. So, in `RequestInstantiateAll(B)`, `RequestInstantiate(B)` returns the stored promise, but __it performs the dependency resolution again in 1-a or later__. So it will call `RequestInstantiateAll(C)` at step 1-b-i-3 again.
8. `RequestInstantiateAll(B)` and `RequestInstantiateAll(C)` are called repeatedly.

This problem comes from that the current algorithm does not store the status of "resolving the dependencies".

To solve this, I would like to propose introducing the new state `ResolveDependencies`. That means, "Now, the instantiation is done, so the  entry is ready to resolve the dependencies. And if there's entry.[[ResolveDependencies]] promise and the state is `ResolveDependencies`, we're resolving the dependency of the module just now".

The algorithm becomes the following,

### CommitInstantiated(loader, key)
1. Let _instance_ be Instantiation(_loader_, _optionalInstance_, _source_).
1. ReturnIfAbrupt(_instance_).
1. // TODO: edge case: what if _instance_ is a thenable function?
1. Fulfill _entry_.[[Instantiate]] with _instance_.
1. Let _deps_ be a new empty List.
1. If _instance_ is a Module Record, then:
  1. Assert: _instance_ is a Source Text Module Record.
  1. Set _instance_.[[RegistryEntry]] to _entry_.
[...]

Issue #40, opened 2015-03-30
Updated Feedback


guybedford:
I've just updated our implementation to match the latest changes. Feedback below.
* 6.3.1 Should call requestInstantiateAll to ensure optimal loading (early dependency loading)
* 6.5.2. Reflect.Loader.install - we should probably throw here if `module` is not a module object.
* 4.1.6 Can call commitInstantiated when entry.[[Instantiate]] is undefined. Advisable to merge with 4.1.5 and just call fulfillInstantiate.
* It is possible to use the fulfillX functions at the end of each requestX function instead of repeating between the two: "Set entry.[[state]] to the max of entry.[[state]] and translate/fetch"
* 4.2.4 depEntry not defined when checking _depEntry_.[[State]], worth using a manual call to `ensureRegistered` at the top like all the others.
* 4.2.4 Circular references still don't have a stop condition here. Some type of extra state is needed, perhaps an `INSTANTIATE_ALL` that fits in after instantiate, which then gets marked as `LINK` within instantiateAll before hitting the circular references.

caridy:
> 6.3.1 Should call requestInstantiateAll to ensure optimal loading (early dependency loading)

ok, will do.

> 6.5.2. Reflect.Loader.install - we should probably throw here if module is not a module object.

need more discussion. @dherman should look into this one. It is my understanding that `[[Module]]` could be undefined when installing it.

> 4.1.6 Can call commitInstantiated when entry.[[Instantiate]] is undefined. Advisable to merge with 4.1.5 and just call fulfillInstantiate.

cosmetic, we want to keep having all the fullfill* abstract operations to be called from public api methods (e.g.: loader.prototype.*). let's keep that separation.

> It is possible to use the fulfillX functions at the end of each requestX function instead of repeating between the two: "Set entry.[[state]] to the max of entry.[[state]] and translate/fetch"

same as above.

> 4.2.4 depEntry not defined when checking depEntry.[[State]], worth using a manual call to ensureRegistered at the top like all the others.

551d06abd9997c9bc845889e584d56fd4cede771


> 4.2.4 Circular references still don't have a stop condition here. Some type of extra state is needed, perhaps an INSTANTIATE_ALL that fits in after instantiate, which then gets marked as LINK within instantiateAll before hitting the circular references.

need more discussion. @dherman should look into this one.

PR #94, opened 2015-10-15
add the new pipeline stage "satisfy"

dherman:
…to properly handle cyclic dependency graphs (see #68 and #40).

Also eliminated a couple of out-of-date lines in ExtractDependencies: the TODO is no longer relevant, since promise assimilation will ensure the instance is not a thenable. And there is no \[[Instantiate]] promise to explicitly fulfill; instead, the promises in the ordinary flow of the RequestXXX abstract operations receive the result naturally.


caridy:
Looking great. The only question is about `.uninstall()`, `.cancel()`, and whether `"satisfy"` should be considered in step 8.

dherman:
I'm going to address `.uninstall` and `.cancel` in a separate issue.

Issue #96, opened 2015-10-23
registry refactoring


dherman:
Splitting out the registry from the loader as a separate object with a distinct class was a big step forward for separating concerns. But the registry abstraction is itself still muddled, so now it's time to turn our attention there.

Conceptually, the registry is just a mutable dictionary mapping string keys to modules (in whatever transitory state of loading). The twist is that a module's loading-state is based on promises, and should be managed in an asynchronous way; we don't want to be manipulating promises in ways that dip below the promise abstraction layer (such as fulfilling with final values that are thenable, or observing promises change state mid-task). But there are a number of use cases where we want to be able to synchronously interact with the registry:

* Setting up polyfills in an execution phase that implicitly blocks later ones that depend on the polyfill. For example, installing modules in a polyfill with `<script>/* set up registry */</script> ... <script type="module">import blah from "blah" ...</script>` should allow the second script to depend on the registry modifications.
* Iterating over the registry does not want to interleave with other turns, so it makes sense to be able to synchronously ask for the current set of keys.

This suggests that we should further separate the concerns of *registration* and *loading status*. In particular, the simplest, most general representation of the registry is effectively a map:

```js
%%RegistryPrototype%%.keys() -> Iterator<string>
%%RegistryPrototype%%.values() -> Iterator<RegistryEntry>
%%RegistryPrototype%%.entries() -> Iterator<[string, RegistryEntry]>
%%RegistryPrototype%%.get(key) -> RegistryEntry
%%RegistryPrototype%%.set(key, entry)
%%RegistryPrototype%%.has(key) -> boolean
%%RegistryPrototype%%.delete(key)
```

The loading status of a registry entry, however, is primarily asynchronous and should be where the methods for querying and updating the loading status live:

```js
get %%RegistryEntryPrototype%%.stage -> "fetch" or "translate" or "instantiate" or "satisfy" or "link" or "ready"
get %%RegistryEntryPrototype%%.module -> module object or null
%%RegistryEntryPrototype%%.load(stage) -> promise
%%RegistryEntryPrototype%%.result(stage) -> promise or undefined
%%RegistryEntryPrototype%%.resolve(stage, result)
%%RegistryEntryPrototype%%.reject(stage, error)
%%RegistryEntryPrototype%%.cancel(stage)
```

[...]

PR #97, opened 2015-10-23
fixes issue #96: registry refactor to introduce registry entry objects

caridy:
### HTML Version

https://rawgit.com/caridy/6c921c618d3cc2b162ef/raw/2528a3224b0962a415fefc65565ac51bac2f2ec0/loader-refactor-registry-entry.html

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -832,7 +832,7 @@ RegistryEntry instances are initially created with the internal slots described 
         1. Let _dep_ be _depEntry_.[[Module]].
         1. Set _pair_.[[Value]] to _dep_.
         1. Return _dep_.
-      1. Return the result of transforming RequestSatisfy(_depEntry_.[[Loader]], _depKey_) with a fulfillment handler that, when called with value _depEntry_, runs the following steps:
+      1. Return the result of transforming RequestSatisfy(_depEntry_) with a fulfillment handler that, when called with value _depEntry_, runs the following steps:
         1. Let _dep_ be _depEntry_.[[Module]].
         1. Set _pair_.[[Value]] to _dep_.
         1. Return _dep_.
--- a/index.bs
+++ b/index.bs
@@ -253,7 +253,7 @@ The following steps are taken:
 
 <h4 id="loader-registry">get Loader.prototype.registry</h4>
 
-Loader.prototype.registry is an accessor property whose set accessor function is undefined. Its get accessor function performs the following steps:
+<code>Loader.prototype.registry</code> is an accessor property whose set accessor function is undefined. Its get accessor function performs the following steps:
 
 <emu-alg>
 1. Let _loader_ be *this* value.
@@ -675,7 +675,7 @@ The initial value of ModuleStatus.prototype.constructor is %Registry%.
 
 <h4 id="module-status-stage">get ModuleStatus.prototype.stage</h4>
 
-ModuleStatus.prototype.stage is an accessor property whose set accessor function is undefined. Its get accessor function performs the following steps:
+<code>ModuleStatus.prototype.stage</code> is an accessor property whose set accessor function is undefined. Its get accessor function performs the following steps:
 
 <emu-alg>
 1. Let _entry_ be *this* value.
@@ -686,7 +686,7 @@ ModuleStatus.prototype.stage is an accessor property whose set accessor function
 
 <h4 id="module-status-module">get ModuleStatus.prototype.module</h4>
 
-ModuleStatus.prototype.module is an accessor property whose set accessor function is undefined. Its get accessor function performs the following steps:
+<code>ModuleStatus.prototype.module</code> is an accessor property whose set accessor function is undefined. Its get accessor function performs the following steps:
 
 <emu-alg>
 1. Let _entry_ be *this* value.
@@ -694,6 +694,49 @@ ModuleStatus.prototype.module is an accessor property whose set accessor functio
 1. If _entry_ does not have a [[Module]] internal slot, throw a *TypeError* exception.
 1. Return _entry_.[[Module]].
 </emu-alg>
+
+<h4 id="module-status-error">get ModuleStatus.prototype.error</h4>
+
+<code>ModuleStatus.prototype.error</code> is an accessor property whose set accessor function is undefined. Its get accessor function performs the following steps:
+
+<emu-alg>
+1. Let _entry_ be *this* value.
+1. If Type(_entry_) is not Object, throw a *TypeError* exception.
+1. If _entry_ does not have a [[Error]] internal slot, throw a *TypeError* exception.
+1. Return _entry_.[[Error]].
+</emu-alg>
+
+
+<h4 id="module-status-error">get ModuleStatus.prototype.dependencies</h4>
+
+<code>ModuleStatus.prototype.dependencies</code> is an accessor property whose set accessor function is undefined. Its get accessor function performs the following steps:
+
+<emu-alg>
+1. Let _entry_ be *this* value.
+1. If Type(_entry_) is not Object, throw a *TypeError* exception.
+1. If _entry_ does not have a [[Dependencies]] internal slot, throw a *TypeError* exception.
+1. Let _array_ be ArrayCreate(0).
+1. Let _n_ be 0.
+1. For each _pair_ in _entry_.[[Dependencies]], do:
+  1. Let _O_ be ObjectCreate(%ObjectPrototype%).
+  1. Let _requestNameDesc_ be the PropertyDescriptor{[[Value]]: _pair_.[[RequestName]], [[Writable]]: *false*, [[Enumerable]]: *true*, [[Configurable]]: *false*}.
+  1. Let _requestNameStatus_ be DefinePropertyOrThrow(_O_, "requestName", _requestNameDesc_).
+  1. ReturnIfAbrupt(_requestNameStatus_).
+  1. Let _keyDesc_ be the PropertyDescriptor{[[Value]]: _pair_.[[Key]], [[Writable]]: *false*, [[Enumerable]]: *true*, [[Configurable]]: *false*}.
+  1. Let _keyStatus_ be DefinePropertyOrThrow(_O_, "key", _keyDesc_).
+  1. ReturnIfAbrupt(_keyStatus_).
+  1. Let _moduleStatusDesc_ be the PropertyDescriptor{[[Value]]: _pair_.[[ModuleStatus]], [[Writable]]: *false*, [[Enumerable]]: *true*, [[Configurable]]: *false*}.
+  1. Let _moduleStatus_ be DefinePropertyOrThrow(_O_, "entry", _moduleStatusDesc_).
+  1. ReturnIfAbrupt(_moduleStatus_).
+  1. Let _status_ be CreateDataProperty(_array_, ToString(_n_), _O_).
+  1. ReturnIfAbrupt(_status_).
+  1. Increment _n_ by 1.
+1. Return _array_.
+</emu-alg>
+
+<emu-note>
+  The dependencies accessor provides access to the dependency graph, exposing each dependency's requestName, key and entry reference.
+</emu-note>
 
 <h4 id="module-status-prototype-load">ModuleStatus.prototype.load(stage)</h4>
 
@@ -807,8 +850,8 @@ ModuleStatus instances are initially created with the internal slots described i
   </tr>
   <tr>
     <td>\[[Dependencies]]</td>
-    <td>List of pairs of String and (<a href="#module-status">module status</a> or <code>undefined</code>), or <code>undefined</code>.</td>
-    <td>Table mapping unresolved names to their resolved modules.</td>
+    <td>List of Records of the form {\[[RequestName]]: String, \[[Key]]: String, \[[ModuleStatus]]: Module Status} .</td>
+    <td>Table mapping unresolved names to their resolved key and module status entries.</td>
   </tr>
   <tr>
     <td>\[[Module]]</td>
@@ -856,10 +899,15 @@ ModuleStatus instances are initially created with the internal slots described i
   1. Assert: _instance_ is a Source Text Module Record.
   1. Set _instance_.[[ModuleStatus]] to _entry_.
   1. For each _dep_ in _instance_.[[RequestedModules]], do:
-    1. Append the record { [[Key]]: _dep_, [[Value]]: *undefined* } to _deps_.
+    1. Append the record { [[RequestName]]: _dep_, [[Key]]: *undefined*, [[ModuleStatus]]: *undefined* } to _deps_.
 1. Set _entry_.[[Dependencies]] to _deps_.
 1. Set _entry_.[[Module]] to _instance_.
 </emu-alg>
+
+<emu-note>
+  [[Key]] wil identifies the dependency module status in the entry's corresponding loader.
+</emu-note>
+
 
 <h4 id="instantiation" aoid="Instantiation">Instantiation(loader, result, source)</h4>
 
@@ -937,15 +985,13 @@ ModuleStatus instances are initially created with the internal slots described i
   1. For each _pair_ in _entry_.[[Dependencies]], do:
     1. Let _p_ be the result of transforming Resolve(_loader_, _pair_.[[Key]], _key_) with a fulfillment handler that, when called with value _depKey_, runs the following steps:
       1. Let _depEntry_ be EnsureRegistered(_entry_.[[Loader]], _depKey_).
+      1. Let _pair_.[[Key]] to _depKey_.
+      1. Let _pair_.[[ModuleStatus]] to _depEntry_.
       1. Let _currentStageEntry_ be GetCurrentStage(_entry_).
       1. If _currentStageEntry_.[[Stage]] is "ready", then:
-        1. Let _dep_ be _depEntry_.[[Module]].
-        1. Set _pair_.[[Value]] to _dep_.
-        1. Return _dep_.
+        1. Return _depEntry_.[[Module]].
       1. Return the result of transforming RequestSatisfy(_depEntry_) with a fulfillment handler that, when called with value _depEntry_, runs the following steps:
-        1. Let _dep_ be _depEntry_.[[Module]].
-        1. Set _pair_.[[Value]] to _dep_.
-        1. Return _dep_.
+        1. Return _depEntry_.[[Module]].
     1. Append _p_ to _depLoads_.
   1. Let _p_ be the result of waiting for all _depLoads_.
   1. Return the result of transforming _p_ with a fulfillment handler that, when called, runs the following steps:
@@ -954,6 +1000,10 @@ ModuleStatus instances are initially created with the internal slots described i
 1. Set _satisfyStageEntry_.[[Result]] to _p_.
 1. Return _p_.
 </emu-alg>
+
+<emu-note>
+  This abstract operation completes the [[Dependencies]] registry per entry keeping backpointers to each dependency entry and the corresponding keys.
+</emu-note>
 
 <h4 id="request-link" aoid="RequestLink">RequestLink(entry)</h4>
 
@@ -995,12 +1045,12 @@ The modules spec should only invoke this operation from methods of Source Text M
 1. Let _entry_ be _module_.[[ModuleStatus]].
 1. Let _currentStageEntry_ be GetCurrentStage(_entry_).
 1. Assert: _currentStageEntry_.[[Stage]] is in "link" or "ready" stage.
-1. Let _pair_ be the pair in _entry_.[[Dependencies]] such that _pair_.[[Key]] is equal to _requestName_.
+1. Let _pair_ be the pair in _entry_.[[Dependencies]] such that _pair_.[[RequestName]] is equal to _requestName_.
 1. Assert: _pair_ is defined.
-1. Let _dep_ be _pair_.[[Value]].
-1. Let _depStageEntry_ be GetCurrentStage(_dep_).
+1. Let _depEntry_ be _pair_.[[ModuleStatus]].
+1. Let _depStageEntry_ be GetCurrentStage(_depEntry_).
 1. Assert: _depStageEntry_.[[Stage]] is equal "link" or "ready" stage.
-1. Return _dep_.[[Module]].
+1. Return _depEntry_.[[Module]].
 </emu-alg>
 
 <h3 id="linking">Linking</h3>
@@ -1044,8 +1094,8 @@ The modules spec should only invoke this operation from methods of Source Text M
 1. If _entry_ is already in _result_, then return *undefined*.
 1. Append _entry_ to _result_.
 1. For each _pair_ in _entry_.[[Dependencies]], do:
-  1. Assert: _pair_.[[Value]] is defined.
-  1. Call ComputeDependencyGraph(_pair_.[[Value]], _result_).
+  1. Assert: _pair_.[[ModuleStatus]] is defined.
+  1. Call ComputeDependencyGraph(_pair_.[[ModuleStatus]], _result_).
 1. Return *undefined*.
 </emu-alg>
 
--- a/index.bs
+++ b/index.bs
@@ -242,10 +242,13 @@ The following steps are taken:
 <emu-alg>
 1. Let _loader_ be *this* value.
 1. Assert: _loader_ must be an instance of a concrete subclass of Loader.
-1. If _stage_ is *undefined* then let _stage_ be "ready".
+1. If _stage_ is *undefined* then let _stageValue_ be "ready".
+1. Else let _stageValue_ be ToString(_stage_).
+1. If _stageValue_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
+1. If IsValidStageValue(_stageValue_) is *false*, return a promise rejected with a new *RangeError* exception.
 1. Return the result of transforming Resolve(_loader_, _name_, _referrer_) with a fulfillment handler that, when called with argument _key_, runs the following steps:
   1. Let _entry_ be EnsureRegistered(_loader_, _key_).
-  1. Return LoadModule(_entry_, _stage_).
+  1. Return LoadModule(_entry_, _stageValue_).
 </emu-alg>
 
 <h4 id="loader-registry">get Loader.prototype.registry</h4>
@@ -476,12 +479,22 @@ The abstract operation GetCurrentStage with argument <i>entry</i> performs the f
 1. Return the first element of _stages_.
 </emu-alg>
 
+<h4 id="module-status-IsValidStageValue" aoid="GetCurrentStage">IsValidStageValue(stage)</h4>
+
+The abstract operation IsValidStageValue with argument <i>stage</i> performs the following steps:
+
+<emu-alg>
+1. If _stage_ is "fetch", "translate", "instantiate", "satisfy", "link" or "ready", return *true*.
+1. Else return *false*.
+</emu-alg>
+
 <h4 id="module-status-GetStage" aoid="GetStage">GetStage(entry, stage)</h4>
 
 The abstract operation GetStage with arguments <i>entry</i> and <i>stage</i> performs the following steps:
 
 <emu-alg>
 1. If _entry_ is not an initialized ModuleStatus instance, throw a new *TypeError* exception.
+1. Assert: Type(_stage_) is String.
 1. Let _stages_ be _entry_.[[Pipeline]].
 1. For each element _entry_ of _stages_, do
   1. If _entry_.[[Stage]] is equal to _stage_, then
@@ -495,6 +508,7 @@ The abstract operation LoadModule with arguments <i>entry</i> and <i>stage</i> p
 
 <emu-alg>
 1. If _entry_ is not an initialized ModuleStatus instance, return a promise rejected with a new *TypeError* exception.
+1. Assert: Type(_stage_) is String.
 1. If _stage_ is "fetch", then:
   1. Return the result of transforming RequestFetch(_entry_) with a new pass-through promise.
 1. If _stage_ is "translate", then:
@@ -520,6 +534,7 @@ The abstract operation LoadModule with arguments <i>entry</i> and <i>stage</i> p
 <emu-alg>
 1. If Type(_entry_) is not Object, throw a *TypeError* exception.
 1. If _entry_ does not have a [[Pipeline]] internal slot, throw a *TypeError* exception.
+1. Assert: Type(_stage_) is String.
 1. Let _pipeline_ be _entry_.[[Pipeline]].
 1. Let _stageEntry_ be GetStage(_entry_, _stage_).
 1. If _stageEntry_ is not *undefined*, then
@@ -686,8 +701,11 @@ The following steps are taken:
 
 <emu-alg>
 1. Let _entry_ be *this* value.
-1. If _stage_ is *undefined* then let _stage_ be "fetch".
-1. Return LoadModule(_entry_, _stage_).
+1. If _stage_ is *undefined* then let _stageValue_ be "fetch".
+1. Else let _stageValue_ be ToString(_stage_).
+1. If _stageValue_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
+1. If IsValidStageValue(_stageValue_) is *false*, return a promise rejected with a new *RangeError* exception.
+1. Return LoadModule(_entry_, _stageValue_).
 </emu-alg>
 
 <h4 id="module-status-prototype-result">ModuleStatus.prototype.result(stage)</h4>
@@ -697,8 +715,10 @@ The following steps are taken:
 <emu-alg>
 1. Let _entry_ be *this* value.
 1. If Type(_entry_) is not Object, return a promise rejected with a new *TypeError* exception.
-1. If _stage_ is not "fetch", "translate", "instantiate", "satisfy", "link" or "ready", return a promise rejected with a new *RangeError* exception.
-1. Let _stageEntry_ be GetStage(_entry_, _stage_).
+1. Let _stageValue_ be ToString(_stage_).
+1. If _stageValue_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
+1. If IsValidStageValue(_stageValue_) is *false*, return a promise rejected with a new *RangeError* exception.
+1. Let _stageEntry_ be GetStage(_entry_, _stageValue_).
 1. If _stageEntry_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
 1. If _stageEntry_ is *undefined*, return a promise resolved with *undefined*.
 1. If _stageEntry_.[[Result]] is *undefined*, return a promise resolved with *undefined*.
@@ -712,16 +732,18 @@ The following steps are taken:
 <emu-alg>
 1. Let _entry_ be *this* value.
 1. If Type(_entry_) is not Object, throw a new *TypeError*.
-1. If _stage_ is not "fetch", "translate", "instantiate", "satisfy", "link" or "ready", return a promise rejected with a new *RangeError* exception.
+1. Let _stageValue_ be ToString(_stage_).
+1. If _stageValue_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
+1. If IsValidStageValue(_stageValue_) is *false*, return a promise rejected with a new *RangeError* exception.
 1. Let _p0_ be the result of transforming _result_ with a new pass-through promise.
 1. Let _p1_ be the result of transforming _p0_ with a fulfillment handler that, when called with argument _value_, runs the following steps:
-  1. Let _stageEntry_ be GetStage(_entry_, _stage_).
+  1. Let _stageEntry_ be GetStage(_entry_, _stageValue_).
   1. If _stageEntry_ is *undefined*, throw a new *TypeError*.
   1. If _stageEntry_.[[Result]] is *undefined*, then
     1. Set _stageEntry_.[[Result]] to a promise resolved with _value_.
   1. Else
     1. Fulfill _stageEntry_.[[Result]] with _value_.
-  1. UpgradeToStage(_entry_, _stage_).
+  1. UpgradeToStage(_entry_, _stageValue_).
   1. Return _value_.
 1. Return _p0_.
 </emu-alg>
@@ -733,16 +755,18 @@ The following steps are taken:
 <emu-alg>
 1. Let _entry_ be *this* value.
 1. If Type(_entry_) is not Object, throw a new *TypeError*.
-1. If _stage_ is not "fetch", "translate", "instantiate", "satisfy", "link" or "ready", return a promise rejected with a new *RangeError* exception.
+1. Let _stageValue_ be ToString(_stage_).
+1. If _stageValue_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
+1. If IsValidStageValue(_stageValue_) is *false*, return a promise rejected with a new *RangeError* exception.
 1. Let _p0_ be the result of transforming _error_ with a new pass-through promise.
 1. Let _p1_ be the result of transforming _p0_ with a fulfillment handler that, when called with argument _value_, runs the following steps:
-  1. Let _stageEntry_ be GetStage(_entry_, _stage_).
+  1. Let _stageEntry_ be GetStage(_entry_, _stageValue_).
   1. If _stageEntry_ is *undefined*, throw a new *TypeError*.
   1. If _stageEntry_.[[Result]] is *undefined*, then
     1. Set _stageEntry_.[[Result]] to a promise rejected with _value_.
   1. Else
     1. Reject _stageEntry_.[[Result]] with _value_.
-  1. UpgradeToStage(_entry_, _stage_).
+  1. UpgradeToStage(_entry_, _stageValue_).
   1. Return _value_.
 1. Return _p0_.
 </emu-alg>
--- a/index.bs
+++ b/index.bs
@@ -152,7 +152,7 @@ mean the same thing as:
 
 <emu-alg>
 1. Let _desc_ be a new PropertyDescriptor record {[[Value]]: _value_, [[Writable]]: *true*, [[Enumerable]]: *true*, [[Configurable]]: *true*}.
-1. Return the result of calling OrdinaryDefineOwnProperty(_obj_, _name_, _desc_).
+1. Return ? OrdinaryDefineOwnProperty(_obj_, _name_, _desc_).
 </emu-alg>
 
 <h3 id="common-operations">Built-in Function Objects</h3>
@@ -168,8 +168,7 @@ We follow the ECMA-262 <a href="http://www.ecma-international.org/ecma-262/6.0/i
 The abstract operation CreateRegistry with no arguments performs the following steps:
 
 <emu-alg>
-1. Let _O_ be OrdinaryCreateFromConstructor(Registry, "%RegistryPrototype%", «[[RegistryMap]]» ).
-1. ReturnIfAbrupt(_O_).
+1. Let _O_ be ? OrdinaryCreateFromConstructor(Registry, "%RegistryPrototype%", «[[RegistryMap]]» ).
 1. Return _O_.
 </emu-alg>
 
@@ -189,8 +188,7 @@ When Loader is called with no arguments, the following steps are taken:
 
 <emu-alg>
 1. If NewTarget is *undefined*, then throw a *TypeError* exception.
-1. Let _O_ be OrdinaryCreateFromConstructor(NewTarget, "Loader.prototype").
-1. ReturnIfAbrupt(_O_).
+1. Let _O_ be ? OrdinaryCreateFromConstructor(NewTarget, "Loader.prototype").
 1. Set _O_’s [[Registry]] internal slot to CreateRegistry().
 1. Return _O_.
 </emu-alg>
@@ -219,7 +217,8 @@ The following steps are taken:
 
 <emu-alg>
 1. Let _loader_ be *this* value.
-1. Assert: _loader_ must be an instance of a concrete subclass of Loader.
+1. If Type(_loader_) is not Object, throw a *TypeError* exception.
+1. If _loader_ does not have all of the internal slots of a Loader Instance (<a href="#loader-internal-slots">3.5</a>), throw a *TypeError* exception.
 1. Return the result of transforming Resolve(_loader_, _name_, _referrer_) with a fulfillment handler that, when called with argument _key_, runs the following steps:
   1. Let _entry_ be EnsureRegistered(_loader_, _key_).
   1. Return LoadModule(_entry_, "ready").
@@ -231,7 +230,8 @@ The following steps are taken:
 
 <emu-alg>
 1. Let _loader_ be *this* value.
-1. Assert: _loader_ must be an instance of a concrete subclass of Loader.
+1. If Type(_loader_) is not Object, throw a *TypeError* exception.
+1. If _loader_ does not have all of the internal slots of a Loader Instance (<a href="#loader-internal-slots">3.5</a>), throw a *TypeError* exception.
 1. Return Resolve(_loader_, _name_, _referrer_).
 </emu-alg>
 
@@ -241,10 +241,11 @@ The following steps are taken:
 
 <emu-alg>
 1. Let _loader_ be *this* value.
-1. Assert: _loader_ must be an instance of a concrete subclass of Loader.
+1. If Type(_loader_) is not Object, throw a *TypeError* exception.
+1. If _loader_ does not have all of the internal slots of a Loader Instance (<a href="#loader-internal-slots">3.5</a>), throw a *TypeError* exception.
 1. If _stage_ is *undefined* then let _stageValue_ be "ready".
 1. Else let _stageValue_ be ToString(_stage_).
-1. If _stageValue_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
+1. RejectIfAbrupt(_stageValue_).
 1. If IsValidStageValue(_stageValue_) is *false*, return a promise rejected with a new *RangeError* exception.
 1. Return the result of transforming Resolve(_loader_, _name_, _referrer_) with a fulfillment handler that, when called with argument _key_, runs the following steps:
   1. Let _entry_ be EnsureRegistered(_loader_, _key_).
@@ -257,6 +258,8 @@ The following steps are taken:
 
 <emu-alg>
 1. Let _loader_ be *this* value.
+1. If Type(_loader_) is not Object, throw a *TypeError* exception.
+1. If _loader_ does not have all of the internal slots of a Loader Instance (<a href="#loader-internal-slots">3.5</a>), throw a *TypeError* exception.
 1. Return _loader_.[[Registry]].
 </emu-alg>
 
@@ -304,8 +307,7 @@ When Registry is called with no arguments, the following steps are taken:
 
 <emu-alg>
 1. If NewTarget is *undefined*, then throw a *TypeError* exception.
-1. Let _O_ be OrdinaryCreateFromConstructor(NewTarget, "%RegistryPrototype%", «[[RegistryMap]]» ).
-1. ReturnIfAbrupt(_O_).
+1. Let _O_ be ? OrdinaryCreateFromConstructor(NewTarget, "%RegistryPrototype%", «[[RegistryMap]]» ).
 1. Let _M_ be ObjectCreate(%MapIteratorPrototype%, «[[Map]], [[MapNextIndex]], [[MapIterationKind]]»).
 1. Set _O_’s [[RegistryMap]] internal slot to _M_.
 1. Return _O_.
@@ -346,6 +348,7 @@ The following steps are taken:
 <emu-alg>
 1. Let _registry_ be *this* value.
 1. If Type(_registry_) is not Object, throw a *TypeError* exception.
+1. If _registry_ does not have all of the internal slots of a Registry Instance (<a href="#registry-internal-slots">4.4</a>), throw a *TypeError* exception.
 1. Let _M_ be _registry_.[[RegistryMap]].
 1. Return CreateMapIterator(_M_, "key+value").
 </emu-alg>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -171,7 +171,7 @@ When Reflect.Loader is called with no arguments, the following steps are taken:
 1. If NewTarget is *undefined*, then throw a *TypeError* exception.
 1. Let _O_ be OrdinaryCreateFromConstructor(NewTarget, "Reflect.Loader.prototype").
 1. ReturnIfAbrupt(_O_).
-1. Let _registry_ to a new Registry(_O_).
+1. Let _registry_ to a new Registry().
```
caridy: registry doesn't need a backpointer to loader anymore.

index.bs
```diff
@@ -135,8 +135,8 @@ Algorithm steps that say
 mean the same thing as:
 
 <emu-alg>
-1. If _x_ is an abrupt completion, then return a promise rejected with _x_.[[value]].
-1. Else if _x_ is a Completion Record, then let _x_ be _x_.[[value]].
+1. If _x_ is an abrupt completion, then return a promise rejected with _x_.[[Value]].
```
caridy: `[[value]]` vs `[[Value]]`, resolved by using `[[Value]]` and `[[Key]]` everywhere.

index.bs
```diff
@@ -225,27 +225,11 @@ The following steps are taken:
...
-    1. Return the result of transforming RequestSatisfy(_loader_, _key_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
-      1. If _entry_.[[Module]] is a Function object, return _entry_.[[Module]].
-      1. Return *undefined*.
-  1. If _stage_ is "link", then:
-    1. Return the result of transforming RequestLink(_loader_, _key_) with a fulfillment handler that returns *undefined*.
-  1. If _stage_ is "ready", then:
-    1. Return the result of transforming RequestReady(_loader_, _key_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
-      1. Return GetModuleNamespace(_entry_.[[Module]]).
-  1. Throw a new *TypeError*.
+  1. Let _entry_ be EnsureRegistered(_loader_, _key_).
```
caridy: `loader.load()` is now almost identical to `loader.import()`, they both rely on `entry.load()` now.

caridy: In fact, `loader.load('foo')`, `loader.import('foo')` and `entryForFoo.load('ready')` are all equivalent in behavior and resolution.

index.bs
```diff
@@ -225,27 +225,11 @@ The following steps are taken:
 1. If Type(_loader_) is not Object, throw a *TypeError* exception.
 1. If _stage_ is *undefined* then let _stage_ be "ready".
 1. Return the result of transforming Resolve(_loader_, _name_, _referrer_) with a fulfillment handler that, when called with argument _key_, runs the following steps:
-  1. If _stage_ is "fetch", then:
-    1. Return the result of transforming RequestFetch(_loader_, _key_) with a new pass-through promise.
```
caridy: this block is now moved into `entry.load()`

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
     <td>An object</td>
-    <td>An instance of Registry (<a href="#registry">4</a>).</td>
+    <td>An instance of <a href="#registry">Registry</a>.</td>
   </tr>
 </table>
 
 <h2 id="registry">Registry Objects</h2>
 
-<h3 id="registry-abstract-operations">Abstract Operations for Registry Objects</h3>
+<h3 id="registry-constructor">The Registry Constructor</h3>
```
caridy: diff quirk, abstract operations were moved into the new section under Registry Entry Objects.

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
+<h4 id="registry-prototype-@@iterator">Registry.prototype[ @@iterator ]()</h4>
+
+When the @@iterator method is called it returns an Iterator object (ES2015 25.1.1.2) that iterates over the registry entries of the \[[Entries]] internal slot, returning each entry as a key value pair. The following steps are taken:
+
+<emu-alg>
+1. Let _registry_ be *this* value.
+1. If Type(_registry_) is not Object, throw a *TypeError* exception.
+1. If _registry_ does not have a [[Entries]] internal slot, throw a *TypeError* exception.
+1. Let _entries_ be a new List.
+1. For each _pair_ in _registry_.[[Entries]], do:
```
caridy: Small change in this method, it now takes a snapshot of the registry, to produce the iterator. I'm not very happy with this, and I hope we can use some abtraction from 262 to do so.

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
+
+The following steps are taken:
+
+<emu-alg>
+1. Let _registry_ be *this* value.
+1. If Type(_registry_) is not Object, throw a *TypeError* exception.
+1. If _registry_ does not have a [[Entries]] internal slot, throw a *TypeError* exception.
+1. Assert: _entry_ is an initialized RegistryEntry instance.
+1. Let _entries_ be _registry_.[[Entries]].
+1. Let _pair_ be the entry in _entries_ such that _pair_.[[Key]] is equal to _key_.
```
caridy: throwing if the `[[Key]]` already exists.

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
+1. If _pair_ exists, then throw a new *TypeError*.
+1. Append { [[Key]]: _key_, [[Value]]: _entry_ } to _entries_.
+1. Return _registry_.
```
caridy: chaining to mimic a Map

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
+
+<emu-alg>
+1. Let _registry_ be *this* value.
+1. If Type(_registry_) is not Object, throw a *TypeError* exception.
+1. If _registry_ does not have a [[Entries]] internal slot, throw a *TypeError* exception.
+1. Let _entries_ be _registry_.[[Entries]].
+1. Let _pair_ be the entry in _entries_ such that _pair_.[[Key]] is equal to _key_.
+1. If _pair_ does not exist, then throw a new *TypeError*.
+1. Remove _pair_ from _entries_.
+1. Return *true*.
```
caridy: true when success, to mimic a Map

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
+    </tr>
+  </thead>
+  <tr>
+    <td>\[[Entries]]</td>
+    <td>List of pairs of String and <a href="#registry-entry">registry entry</a>.</td>
+    <td>The registry of installed modules.</td>
+  </tr>
+</table>
+
+<h2 id="registry-entry">Registry Entry Objects</h2>
```
caridy: This is the new section.

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
+
+<h3 id="registry-entry-abstract-operations">Abstract Operations for Registry Entry Objects</h3>
+
+<h4 id="registry-entry-GetCurrentStage" aoid="GetCurrentStage">GetCurrentStage(entry)</h4>
 
 The abstract operation GetCurrentStage with argument entry performs the following steps:
 
 <emu-alg>
 1. If Type(_entry_) is not Object, throw a *TypeError* exception.
+1. If _entry_ does not have a [[Pipeline]] internal slot, throw a *TypeError* exception.
```
caridy: I have started adding more checks everywhere to match 262 style, in preparation for the lifting.

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-1. If _entry_.[[Error]] is *nothing*, then:
-  1. Call SimpleDefine(_result_, "error", *null*).
-1. Else:
-  1. Let _opt_ be CreateObject().
-  1. Call SimpleDefine(_opt_, "value", _entry_.[[Error]]).
-  1. Call SimpleDefine(_result_, "error", _opt_).
-1. Return _result_.
-</emu-alg>
-
-A <dfn id="registry-entry">registry entry</dfn> is a record with the following fields:
```
caridy: this was moved down to the properties of the registry entry instance, under internal slot subsection.

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-<h4 id="new-registry" aoid="Registry">Registry(loader)</h4>
+<h4 id="new-registry-entry" aoid="RegistryEntry">RegistryEntry(loader)</h4>
 
 When Registry is called with argument <i>loader</i>, the following steps are taken:
 
 <emu-alg>
 1. If NewTarget is *undefined*, then throw a *TypeError* exception.
 1. If Type(_loader_) is not Object, throw a *TypeError* exception.
-1. Let _O_ be OrdinaryCreateFromConstructor(NewTarget, "%RegistryPrototype%", «[[RegistryData]]» ).
+1. Let _O_ be OrdinaryCreateFromConstructor(NewTarget, "%RegistryPrototype%", «[[Loader]], [[Pipeline]], [[Module]], [[Metadata]], [[Dependencies]], [[Error]]» ).
```
caridy: back pointer to the `[[Loader]]` that the entry is bound to.

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-1. Let _registry_ be *this* value.
-1. Return GetRegistryEntry(_registry_, _key_).
+1. Let _entry_ be *this* value.
+1. If Type(_entry_) is not Object, throw a *TypeError* exception.
+1. Let _stageEntry_ be GetCurrentStage(_entry_).
+1. Return _stageEntry_.[[Stage]].
 </emu-alg>
 
-<h4 id="registry-prototype-install">Registry.prototype.install(key, module)</h4>
+<h4 id="registry-entry-module">get RegistryEntry.prototype.module</h4>
```
caridy: get %%RegistryEntryPrototype%%.module -> module object or null

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
+<h3 id="registry-entry-prototype-object">Properties of the Registry Entry Prototype Object</h3>
 
-<h4 id="registry-prototype-constructor">Registry.prototype.constructor</h4>
+<h4 id="registry-entry-prototype-constructor">RegistryEntry.prototype.constructor</h4>
 
-The initial value of Registry.prototype.constructor is %Registry%.
+The initial value of RegistryEntry.prototype.constructor is %Registry%.
 
-<h4 id="registry-prototype-@@iterator">Registry.prototype[ @@iterator ]()</h4>
+<h4 id="registry-entry-stage">get RegistryEntry.prototype.stage</h4>
```
caridy: `get %%RegistryEntryPrototype%%.stage -> "fetch" or "translate" or "instantiate" or "satisfy" or "link" or "ready"`

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-1. Let _entry_ be a new registry entry record { [[Key]]: _key_, [[Pipeline]]: _pipeline_, [[Metadata]]: *undefined*, [[Dependencies]]: *undefined*, [[Module]]: _module_ }.
-1. Append { [[key]]: _key_, [[value]]: _entry_ } to _entries_.
+1. Let _entry_ be *this* value.
+1. If Type(_entry_) is not Object, throw a *TypeError* exception.
+1. If _entry_ does not have a [[Module]] internal slot, throw a *TypeError* exception.
+1. Return _entry_.[[Module]].
 </emu-alg>
 
-<h4 id="registry-prototype-uninstall">Registry.prototype.uninstall(key)</h4>
+<h4 id="registry-entry-prototype-load">RegistryEntry.prototype.load(stage)</h4>
```
caridy: this one does the bulk of the job now.

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-1. Let _registry_ be *this* value.
-1. If Type(_registry_) is not Object, throw a *TypeError* exception.
-1. Let _entries_ be _registry_.[[RegistryData]].
-1. Let _pair_ be the entry in _entries_ such that _pair_.[[key]] is equal to _key_.
-1. If _pair_ does not exist, then throw a new *TypeError*.
-1. Let _entry_ be _pair_.[[value]].
-1. Let _stageEntry_ be GetCurrentStage(_entry_).
-1. If _stageEntry_.[[Stage]] is not "link" or "ready", throw a new *TypeError*.
-1. Remove _pair_ from _entries_.
+1. Let _entry_ be *this* value.
```
caridy: this block was previous under `Reflect.Loader.prototype.load`

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
+1. If _stage_ is "link", then:
+  1. Return the result of transforming RequestLink(_entry_) with a fulfillment handler that returns *undefined*.
+1. If _stage_ is "ready" or *undefined*, then:
+  1. Return the result of transforming RequestReady(_entry_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
+    1. Return GetModuleNamespace(_entry_.[[Module]]).
+1. Return a promise rejected with a new *RangeError* exception.
 </emu-alg>
 
-<h4 id="registry-prototype-cancel">Registry.prototype.cancel(key)</h4>
+<h4 id="registry-entry-prototype-result">RegistryEntry.prototype.result(stage)</h4>
```
caridy: `%%RegistryEntryPrototype%%.result(stage) -> promise or resolves to undefined`

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
+1. If _stage_ is not "fetch", "translate", "instantiate", "satisfy", "link" or "ready", return a promise rejected with a new *RangeError* exception.
+1. Let _stageEntry_ be GetStage(_entry_, _stage_).
+1. If _stageEntry_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
+1. If _stageEntry_ is *undefined*, return a promise resolved with *undefined*.
+1. If _stageEntry_.[[Result]] is *undefined*, return a promise resolved with *undefined*.
+1. Return the result of transforming _stageEntry_.[[Result]] with a new pass-through promise.
 </emu-alg>
 
-<h4 id="registry-prototype-provide">Registry.prototype.provide(key, stage, value)</h4>
+<h4 id="registry-entry-prototype-provide">RegistryEntry.prototype.resolve(stage, result)</h4>
```
caridy: `%%RegistryEntryPrototype%%.resolve(stage, result)` this one promises on `result`

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
+  1. Return _value_.
+1. Return _p0_.
 </emu-alg>
 
-<emu-note>
-  Any stage other than fetch, translate and instantiate is considered internal.
-</emu-note>
-
-<h4 id="registry-prototype-error">Registry.prototype.error(key, stage, value)</h4>
+<h4 id="registry-entry-prototype-reject">RegistryEntry.prototype.reject(stage, error)</h4>
```
caridy: `%%RegistryEntryPrototype%%.reject(stage, error)` this one promises on error

index.bs
```diff
@@ -666,17 +738,17 @@ Registry instances are initially created with the internal slots described in th
 1. Return the result of promise-calling _hook_(_name_, _referrer_).
 </emu-alg>
 
-<h4 id="extract-dependencies" aoid="ExtractDependencies">ExtractDependencies(loader, entry, optionalInstance, source)</h4>
+<h4 id="extract-dependencies" aoid="ExtractDependencies">ExtractDependencies(entry, optionalInstance, source)</h4>
```
caridy: extracting dependencies on an entry, not longer on a `key`, which means synchronous modifications of the registry will not affect this flow.

index.bs
```diff
@@ -692,15 +764,14 @@ Registry instances are initially created with the internal slots described in th
 
 <h3 id="loading-operations">Loading Operations</h3>
 
-<h4 id="request-fetch" aoid="RequestFetch">RequestFetch(loader, key)</h4>
+<h4 id="request-fetch" aoid="RequestFetch">RequestFetch(entry)</h4>
```
caridy: all `Request*()` abstract operations are now using `entry` instead of `key`, and the internal `[[Loader]]` backpointer on every entry.

index.bs
```diff
@@ -749,26 +816,25 @@ Registry instances are initially created with the internal slots described in th
...
+<h4 id="request-satisfy" aoid="RequestSatisfy">RequestSatisfy(entry)</h4>
 
 <emu-alg>
-1. Let _entry_ be EnsureRegistered(_loader_, _key_).
-1. Let _satisfyStageEntry_ be GetStage(_entry_, "satisfy").
+1. Let _satisfyStageEntry_ be GetStage(_entry_.[[Loader]], "satisfy").
 1. If _satisfyStageEntry_ is *undefined*, return a promise resolved with *undefined*.
 1. If _satisfyStageEntry_.[[Result]] is not *undefined*, return _satisfyStageEntry_.[[Result]].
-1. Let _p_ be the result of transforming RequestInstantiate(_loader_, _key_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
+1. Let _p_ be the result of transforming RequestInstantiate(_entry_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
```
caridy: not happy with _entry_ and _entry_ from the resolved promise... suggestions?

index.bs
```diff
@@ -749,26 +816,25 @@ Registry instances are initially created with the internal slots described in th
...
   1. Let _depLoads_ be a new empty List.
   1. For each _pair_ in _entry_.[[Dependencies]], do:
-    1. Let _p_ be the result of transforming Resolve(_loader_, _pair_.[[key]], _key_) with a fulfillment handler that, when called with value _depKey_, runs the following steps:
-      1. Let _depEntry_ be EnsureRegistered(_loader_, _depKey_).
+    1. Let _p_ be the result of transforming Resolve(_loader_, _pair_.[[Key]], _key_) with a fulfillment handler that, when called with value _depKey_, runs the following steps:
+      1. Let _depEntry_ be EnsureRegistered(_entry_.[[Loader]], _depKey_).
```
caridy: depKey will be resolved on the dependant loader, to support entries bound to different loaders.

index.bs
```diff
@@ -779,27 +845,25 @@ Registry instances are initially created with the internal slots described in th
...
 
-<h4 id="request-link" aoid="RequestLink">RequestLink(loader, key)</h4>
+<h4 id="request-link" aoid="RequestLink">RequestLink(entry)</h4>
 
 <emu-alg>
-1. Let _entry_ be EnsureRegistered(_loader_, _key_).
 1. Let _linkStageEntry_ be GetStage(_entry_, "link").
 1. If _linkStageEntry_ is *undefined*, return a promise resolved with *undefined*.
-1. Return the result of transforming RequestSatisfy(_loader_, _key_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
+1. Return the result of transforming RequestSatisfy(_entry_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
```
caridy: same as above, entry and entry after resolving the promise, while are suppose to be the same...

index.bs
```diff
@@ -749,26 +816,25 @@ Registry instances are initially created with the internal slots described in th
...
+    1. Let _p_ be the result of transforming Resolve(_loader_, _pair_.[[Key]], _key_) with a fulfillment handler that, when called with value _depKey_, runs the following steps:
+      1. Let _depEntry_ be EnsureRegistered(_entry_.[[Loader]], _depKey_).
       1. Let _currentStageEntry_ be GetCurrentStage(_entry_).
       1. If _currentStageEntry_.[[Stage]] is "ready", then:
         1. Let _dep_ be _depEntry_.[[Module]].
-        1. Set _pair_.[[value]] to _dep_.
+        1. Set _pair_.[[Value]] to _dep_.
         1. Return _dep_.
-      1. Return the result of transforming RequestSatisfy(_loader_, _depKey_) with a fulfillment handler that, when called with value _depEntry_, runs the following steps:
+      1. Return the result of transforming RequestSatisfy(_depEntry_.[[Loader]], _depKey_) with a fulfillment handler that, when called with value _depEntry_, runs the following steps:
```
caridy: ops, this should be fixed to use _depEntry_

index.bs
caridy: @bterlson do you have a recommendation here?

index.bs
dherman: Let _registry_ **be**

index.bs
```diff
@@ -225,27 +225,11 @@ The following steps are taken:
...
-      1. If _entry_.[[Module]] is a Function object, return _entry_.[[Module]].
-      1. Return *undefined*.
-  1. If _stage_ is "link", then:
-    1. Return the result of transforming RequestLink(_loader_, _key_) with a fulfillment handler that returns *undefined*.
-  1. If _stage_ is "ready", then:
-    1. Return the result of transforming RequestReady(_loader_, _key_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
-      1. Return GetModuleNamespace(_entry_.[[Module]]).
-  1. Throw a new *TypeError*.
+  1. Let _entry_ be EnsureRegistered(_loader_, _key_).
+  1. Return _entry_.load(_stage_).
```
dherman: I don't think we want to call the public .load method on the entry, do we? Don't we want to call an abstract operation (so it can't be overwritten in userland)?

caridy: oh, good catch! I will abstract that out.

index.bs
dherman: Awesome, thanks -- the "is not Object" thing was always insufficient anyway.

index.bs
dherman: I think now that it's just basically a map there's no point in this restriction. It's equally expressive whether we require explicit delete or not, but this is needlessly inconsistent with Map and with object-dictionaries.

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
+    <td>An instance of <a href="#registry">Registry</a>.</td>
   </tr>
 </table>
 
 <h2 id="registry">Registry Objects</h2>
 
-<h3 id="registry-abstract-operations">Abstract Operations for Registry Objects</h3>
+<h3 id="registry-constructor">The Registry Constructor</h3>
+
+The Registry constructor is the %Registry% intrinsic object. When called as a constructor it creates and initializes a new %Registry% object. %Registry% is not intended to be called as a function and will throw an exception when called in that manner.
```
dherman: It's not clear to me whether we need to expose the ability to construct new registry objects as a userland function or not. AFAICT the question comes down to subclassing: do subclasses need the ability to specify their own registry, and therefore need the ability to construct new ones (or new instances of a subclass of Registry)?

If they do, then we probably need to have Loader take an argument specifying the registry object. Note that a consequence of this would be that multiple loaders could end up sharing a registry. (This _might_ be fine, even though I've avoided it in the past; every built-in loader in the browser would by default have a distinct registry but you could choose to share registries explicitly if you wanted to.)

But if none of this is necessary, and it's fine for loaders, whether direct instances of Loader or instances of subclasses, to forcibly and non-overrideably get their \[\[Registry]] field automatically set to a new instance of %Registry%, then it doesn't make much sense to me to expose this constructor functionality.

index.bs
caridy: agreed, that's why I have added an open question: should we use a Map instead? There is no backpointer to loader, this is just a map :)

index.bs
caridy: Agreed, the loader sets the [[Registry]] internal slot automatically, whether that is an internal %Registry% or just a Map.

dherman: OK, but what I'm saying is that we should have an abstract CreateRegistry operation that is used by the Loader constructor to create a new Registry object, and the function exposed as `loader.registry.constructor` should just be a placeholder function that always throws (whether called or `new`ed).

index.bs
dherman: It's tempting, but I really think we want the registry to be a type-restricted version of Map: the keys can only be strings, and the values can only be RegistryEntry objects. It'd be nice to reuse some of the methods of `Map.prototype` but if we do that then someone could subvert the type invariant by doing e.g. `Map.prototype.set.call(registry, 666, "zomg")`.

So maybe the closest we can get to reuse is to have an internal \[\[RegistryMap]] property that contains a Map, and have all the %RegistryPrototype% methods do the type checks and then delegate to the map's operations. But it can't just call methods directly on that map, because `Map.prototype` could've been tampered with. So I think the operations would have to call out to "the original meaning of `Map.prototype.XXX`" etc etc.

index.bs
```diff
@@ -201,9 +201,9 @@ The following steps are taken:
 <emu-alg>
 1. Let _loader_ be *this* value.
 1. If Type(_loader_) is not Object, throw a *TypeError* exception.
-1. If _loader_ does not have a [[Registry]] internal slot throw a *TypeError* exception.
 1. Return the result of transforming Resolve(_loader_, _name_, _referrer_) with a fulfillment handler that, when called with argument _key_, runs the following steps:
-  1. Return RequestReady(_loader_, _key_).
+  1. Let _entry_ be EnsureRegistered(_loader_, _key_).
+  1. Return _entry_.load().
```
dherman: I'm torn about whether we want to directly call `.load()`, which could've been tampered with, versus calling some high-integrity abstract operation.

For an internal operation I'd feel strongly about this, but `.import` is a high-level method, so maybe it's fine for it just to be thought of as sugar for `_entry_.load()`.

dherman: This is another place where we're calling `.load()` instead of an abstract operation.

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
+<h3 id="registry-constructor">The Registry Constructor</h3>
+
+The Registry constructor is the %Registry% intrinsic object. When called as a constructor it creates and initializes a new %Registry% object. %Registry% is not intended to be called as a function and will throw an exception when called in that manner.
+
+<h4 id="new-registry" aoid="Registry">Registry()</h4>
+
+When Registry is called with no arguments, the following steps are taken:
+
+<emu-alg>
+1. If NewTarget is *undefined*, then throw a *TypeError* exception.
```
dherman: We should mention in the conventions section that we're using the same convention for NewTarget as ES6 -- it's described in [9.3: Built-in Function Objects](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-built-in-function-objects). We can add a subsection called Built-in Function Objects and say "We follow the ECMA-262 [convention for built-in function objects](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-built-in-function-objects) in which the value of NewTarget in each function argument is *undefined* for \[\[Call]] and the _newTarget_ parameter for \[\[Construct]]."

index.bs
dherman: I think if we just use a map internally as I proposed above, then we can delegate to simply producing iterators for that map (via 262's CreateMapIterator).

Also, we should have not just @@iterator but `keys`, `values`, and `entries`, and @@iterator should produce the same result as `entries`.

index.bs
```diff
@@ -282,30 +266,164 @@ Loader instances are initially created with the internal slots described in the
...
+<table>
+  <thead>
+    <tr>
+      <th>Internal Slot</th>
+      <th>Value Type (<em>non-normative</em>)</th>
+      <th>Description (<em>non-normative</em>)</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>\[[Entries]]</td>
```
dherman: It seems to be 262 style to have the internal property name be prefixed with the container type (e.g. \[\[MapData]]), so we should have \[\[RegistryEntries]] instead of \[\[Entries]].

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
 1. Return _O_.
 </emu-alg>
 
-<h3 id="properties-of-the-registry-constructor">Properties of the Registry Constructor</h3>
+<emu-note>
+  A registry entry is associated to a loader instance at the time of its creation, the [[Loader]] backpointer reflects that.
+</emu-note>
+
+<emu-note>
+  A registry entry can be set into multiple Register instances.
```
dherman: Registry

index.bs
dherman: I want to propose "Module Status Objects" instead of "Registry Entry Objects". To me "status" is the perfect word to express that it's a stateful object representing where we are in the process of loading and initializing a module.

We should also think about a place to expose the constructor. I'd suggest `System.ModuleStatus`. (Since the constructor has to take a loader, it doesn't technically carry any ability to talk the outside world, so it _could_ go in `Reflect`, e.g. `Reflect.Module.Status`, but since `Loader` is namespaced under `System` it makes sense for this constructor to live under `System` as well.)

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-The Registry constructor is the %Registry% intrinsic object. When called as a constructor it creates and initializes a new %Registry% object. %Registry% is not intended to be called as a function and will throw an exception when called in that manner.
-
-<h4 id="new-registry" aoid="Registry">Registry(loader)</h4>
+<h4 id="new-registry-entry" aoid="RegistryEntry">RegistryEntry(loader)</h4>
 
 When Registry is called with argument <i>loader</i>, the following steps are taken:
 
 <emu-alg>
 1. If NewTarget is *undefined*, then throw a *TypeError* exception.
 1. If Type(_loader_) is not Object, throw a *TypeError* exception.
```
dherman: Also check that _loader_ is a Loader object.

dherman: Also check that _loader_ is a Loader object.

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
+
+<emu-note>
+  A registry entry can be set into multiple Register instances.
+</emu-note>
+
+<h3 id="properties-of-the-registry-entry-constructor">Properties of the Registry Entry Constructor</h3>
 
 The value of the \[[Prototype]] internal slot of the Registry constructor is the intrinsic object %FunctionPrototype%.
 
 Besides the internal slots and the length property (whose value is 0), the Registry constructor has the following properties:
```
dherman: Shouldn't the value of `length` be 1 since it expects a single argument (the loader)?

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
+1. Let _pipeline_ be a new List.
+1. Add new stage entry record { [[Stage]]: "fetch", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Add new stage entry record { [[Stage]]: "translate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Add new stage entry record { [[Stage]]: "instantiate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Add new stage entry record { [[Stage]]: "satisfy", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Add new stage entry record { [[Stage]]: "link", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Add new stage entry record { [[Stage]]: "ready", [[Result]]: *undefined* } as a new element of the list _pipeline_.
 1. Set _O_’s [[Loader]] internal slot to _loader_.
+1. Set _O_’s [[Pipeline]] internal slot to _pipeline_.
+1. Set _O_’s [[Module]] internal slot to *undefined*.
```
dherman: Should we use **null** instead of **undefined**, since it's an object ("null is the nonexistent object") and not an arbitrary value ("undefined is the nonexistent value")?

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
+1. If _stage_ is "satisfy", then:
+  1. Return the result of transforming RequestSatisfy(_entry_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
+    1. If _entry_.[[Module]] is a Function object, return _entry_.[[Module]].
+    1. Return *undefined*.
+1. If _stage_ is "link", then:
+  1. Return the result of transforming RequestLink(_entry_) with a fulfillment handler that returns *undefined*.
+1. If _stage_ is "ready" or *undefined*, then:
+  1. Return the result of transforming RequestReady(_entry_) with a fulfillment handler that, when called with argument _entry_, runs the following steps:
+    1. Return GetModuleNamespace(_entry_.[[Module]]).
+1. Return a promise rejected with a new *RangeError* exception.
```
dherman: I guess there's nothing I can do at this point (given the precedent set by Intl) but it's just so so wrong to throw RangeError when validating an argument against a fixed set of constant values.

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-1. If Type(_registry_) is not Object, throw a *TypeError* exception.
-1. Let _entries_ be _registry_.[[RegistryData]].
-1. Let _pair_ be the entry in _entries_ such that _pair_.[[key]] is equal to _key_.
-1. If _pair_ does not exist, then throw a new *TypeError*.
-1. Let _entry_ be _pair_.[[value]].
-1. Let _stageEntry_ be GetCurrentStage(_entry_).
-1. If _stageEntry_.[[Stage]] is not "link" or "ready", throw a new *TypeError*.
-1. Remove _pair_ from _entries_.
+1. Let _entry_ be *this* value.
+1. If _entry_ is not an initialized RegistryEntry instance, then return a promise rejected with a new *TypeError* exception.
```
dherman: We should probably do the equivalent of:
```js
if (typeof stage === 'undefined') {
  stage = 'fetch';
}
try {
    stage = String(stage);
} catch (error) {
    return Promise.reject(error);
}
```

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-1. If Type(_registry_) is not Object, throw a *TypeError* exception.
-1. Let _entries_ be _registry_.[[RegistryData]].
-1. Let _pair_ be the entry in _entries_ such that _pair_.[[key]] is equal to _key_.
-1. If _pair_ does not exist, then throw a new *TypeError*.
-1. Let _entry_ be _pair_.[[value]].
-1. Let _stageEntry_ be GetCurrentStage(_entry_).
-1. If _stageEntry_.[[Stage]] is "link" or "ready", throw a new *TypeError*.
-1. Remove _pair_ from _entries_.
+1. Let _entry_ be *this* value.
+1. If Type(_entry_) is not Object, return a promise rejected with a new *TypeError* exception.
```
dherman: Again, should probably do:
```js
try {
  stage = String(stage);
} catch (error) {
  return Promise.reject(error);
}
```

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-</table>
-
-<h3 id="registry-constructor">The Registry Constructor</h3>
-
-The Registry constructor is the %Registry% intrinsic object. When called as a constructor it creates and initializes a new %Registry% object. %Registry% is not intended to be called as a function and will throw an exception when called in that manner.
-
-<h4 id="new-registry" aoid="Registry">Registry(loader)</h4>
+<h4 id="new-registry-entry" aoid="RegistryEntry">RegistryEntry(loader)</h4>
 
 When Registry is called with argument <i>loader</i>, the following steps are taken:
```
dherman: RegistryEntry, but see my note about renaming it.

index.bs
dherman: I think we should allow an optional second argument, which is required to be a valid module object if it is not **undefined**. This would be the way to synchronously circumvent the asynchronous loading process. So you can synchronously create a module status object for a completely loaded module via `new System.ModuleStatus(loader, module)`, and it will fast-forward directly to the ready state; otherwise it will start out in the initial state and can be advanced asynchronously via `.resolve` and `.reject`.

index.bs
```diff
@@ -387,233 +507,172 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-1. Let _entry_ be EnsureRegistered(_loader_, _key_).
-1. Let _stageEntry_ be GetStage(_entry_, _stage_).
-1. If _stageEntry_ is *undefined*, throw a new *TypeError*.
-1. If _stageEntry_.[[Result]] is *undefined*, then set _entry_.[[Result]] to a promise resolved with _value_.
-1. Else, fulfill _entry_.[[Result]] with _value_.
-1. UpgradeToStage(_entry_, _stage_).
+1. Let _entry_ be *this* value.
+1. If Type(_entry_) is not Object, throw a new *TypeError*.
+1. If _stage_ is not "fetch", "translate", "instantiate", "satisfy", "link" or "ready", return a promise rejected with a new *RangeError* exception.
+1. Let _p0_ be the result of transforming _result_ with a new pass-through promise.
```
dherman: (Is that the standard way of specifying the equivalent of calling `Promise.resolve` on a value?)

index.bs
```diff
@@ -692,15 +764,14 @@ Registry instances are initially created with the internal slots described in th
...
 
 <emu-alg>
-1. Let _entry_ be EnsureRegistered(_loader_, _key_).
 1. Let _fetchStageEntry_ be GetStage(_entry_, "fetch").
 1. If _fetchStageEntry_ is *undefined*, return a promise resolved with *undefined*.
-1. Let _hook_ be GetMethod(_loader_, @@fetch).
+1. Let _hook_ be GetMethod(_entry_.[[Loader]], @@fetch).
 1. // TODO: metadata object
-1. Let _p0_ be the result of promise-calling _hook_(_key_).
+1. Let _p0_ be the result of promise-calling _hook_(_entry_).
```
dherman: OK so this is a major bug: now that there's no key, there's no way for a fetch hook to ever know what to fetch. :)

Now, granted, part of this whole PR is about the fact that there isn't a single unique key for an entry since it can be stored at multiple keys or in multiple registries. But somewhere along the line, we _have_ to tell the fetch hook what we're trying to fetch.

index.bs
caridy: I have that, and delete it because of the metadata, will add it back.

index.bs
caridy: I'm not sure I follow. Is the length related to arguments? I didn't know that, I thought it was `Reflect.Loader.length` for example.

index.bs
caridy: I was following the existing definition for that slot: 

```
<tr>
    <td>\[[Module]]</td>
    <td>Module Record or Function object or <code>undefined</code></td>
    <td>The Module Record if the entry is ready, or a thunk if the entry is delayed; otherwise <code>undefined</code>.</td>
  </tr>
```

index.bs
caridy: Hmm... so?

index.bs
caridy: I have been using two forms:

1. `Let _p0_ be the result of transforming _result_ with a new pass-through promise.`
2. `Let _p0_ be a promise resolved with _result_.`

The main difference here is that sometimes we want to store resolve and reject in some internal slots so we can force to resolve those (at least in the past we were forcing those), this is mostly when we pass the promise around, in the other hand, if the result promise is never exposed, we normally use the second option, which is just Promise.resolve().

index.bs
caridy: done

caridy:
updated https://rawgit.com/caridy/6c921c618d3cc2b162ef/raw/d5e42ed44af0b935233374b1553832b3619c169b/loader-refactor-registry-entry.html

matthewp:
Trying to follow along as best as I can, is it correct that the hooks will now expect a RegistryEntry object?  This worries me a little bit as it's pretty important that some of the hooks, particularly fetch, are idempotent and can be called outside of the context of loading a module.  One example would be a plugin syntax where you might have a key like `some-plugin!http://example.com/module.js` and you need to pass only the part after the ! into fetch.  Just want to make sure that the hooks can still be called this way.

caridy:
@matthewp no much is changing with respect to that I think. In the existing spec, if you use the key `some-plugin!http://example.com/module.js` (produced by the resolve hook), that is what will get stored in the registry, and that is what is is going to be passed to the fetch hook, whatever you do with that, it is in your control, you can't have two keys, or a key and the url in the current spec, this PR doesn't change that.

with respect to the fetch hook, it will now receive an entry (in case you need it), and the key used to create that entry in the first place.

matthewp:
> whatever you do with that, it is in your control, you can't have two keys, or a key and the url in the current spec, this PR doesn't change that.

Yes, my question is more about what does the default System loader do with that.  Meaning if I call

```js
System.loader[@@fetch]("http://example.com/module.js")
```

Before I would expect that I could do this and would get a Promise that would resolve with the source text regardless of whether that key is in the registry or not.  So my question is 1) was I wrong about that before and 2) is this not going to be possible with this change?

caridy:
`@@fetch` is entirely at your disposal, you can replace it, or call it directly if you want, but loader will not know about it, not today, and certainly not with this change. You're just bypassing everything to invoke the fetch hook provided by browsers.

Whether that default implementation uses the entry or not, that is yet to be defined, but you can always do this:

```
let o = new Reflect.Module.Status(System.loader, 'http://example.com/module.js');
o.load('fetch').then((payload) => {...});
```

Which will respect the loader's internal process, calling the fetch hook, and get the entry in the fetching stage.

caridy:
updated: https://rawgit.com/caridy/6c921c618d3cc2b162ef/raw/979589815a63b418c0f0f2283a91e9cf5e37f7a6/loader-refactor-registry-entry.html

With this refactor, `entry.dependencies` returns an array of `{requestName: <String>, key: <String>, entry: <ModuleStatus>}`, so you can walk the dependency graph.

The only missing part is the `entry.[[Error]]`, which is not populated today when an error occur.

guybedford:
@caridy :+1: really great to see `entry.dependencies` being provided here!

index.bs
```diff
@@ -155,114 +155,120 @@ mean the same thing as:
...
 
-When Reflect.Loader is called with no arguments, the following steps are taken:
+<h4 id="new-loader">Loader()</h4>
+
+When Loader is called with no arguments, the following steps are taken:
 
 <emu-alg>
 1. If NewTarget is *undefined*, then throw a *TypeError* exception.
-1. Let _O_ be OrdinaryCreateFromConstructor(NewTarget, "Reflect.Loader.prototype").
+1. Let _O_ be OrdinaryCreateFromConstructor(NewTarget, "Loader.prototype").
```
dherman: Shouldn't this be `"%LoaderPrototype%"` instead of `"Loader.prototype"` or `"Reflect.Loader.prototype"`?

caridy: we dont' have `"%LoaderPrototype%"` anymore. We have changed everything from `Reflect.Loader` to just `Loader` objects. I can bring `"%LoaderPrototype%"`.

index.bs
```diff
@@ -155,114 +155,120 @@ mean the same thing as:
...
-<h4 id="reflect-loader-import">Reflect.Loader.prototype.import(name[, referrer])</h4>
+<h4 id="loader-import">Loader.prototype.import(name[, referrer])</h4>
 
 The following steps are taken:
 
 <emu-alg>
 1. Let _loader_ be *this* value.
-1. If Type(_loader_) is not Object, throw a *TypeError* exception.
-1. If _loader_ does not have a [[Registry]] internal slot throw a *TypeError* exception.
+1. Assert: _loader_ must be an instance of a concrete subclass of Loader.
```
dherman: This isn't what "Assert" means in the 262 spec. It's an internal assertion about the coherence of the spec itself (intuitively, a failed Assert means the JS engine should crash/panic/segfault). A check that results in a JS exception should be done with an explicit *TypeError* exception.

We should tell bterlson that this deserves an abstraction to avoid boilerplate, BTW. But it's separate from Assert.

index.bs
```diff
@@ -155,114 +155,120 @@ mean the same thing as:
...
 
-<h4 id="reflect-load-resolve">Reflect.Loader.prototype.resolve(name[, referrer])</h4>
+<h4 id="loader-load-resolve">Loader.prototype.resolve(name[, referrer])</h4>
 
 The following steps are taken:
 
 <emu-alg>
 1. Let _loader_ be *this* value.
-1. If Type(_loader_) is not Object, throw a *TypeError* exception.
+1. Assert: _loader_ must be an instance of a concrete subclass of Loader.
```
dherman: not an Assert

index.bs
dherman: I'm happy with this. `load` is the more generic name for the more generic/configurable operation; `import` is the intuitive name corresponding to the analogous declarative syntax. And `entryForFoo.load('ready')` depends on the name already being resolved and the entry already being in the registry so it's actually pretty different.

index.bs
```diff
@@ -282,30 +288,213 @@ Loader instances are initially created with the internal slots described in the
...
+
+<h4 id="new-registry" aoid="Registry">Registry()</h4>
 
-The abstract operation GetCurrentStage with argument entry performs the following steps:
+When Registry is called with no arguments, the following steps are taken:
 
 <emu-alg>
-1. If Type(_entry_) is not Object, throw a *TypeError* exception.
+1. If NewTarget is *undefined*, then throw a *TypeError* exception.
+1. Let _O_ be OrdinaryCreateFromConstructor(NewTarget, "%RegistryPrototype%", «[[RegistryMap]]» ).
```
dherman: This should probably just call CreateRegistry() to avoid the spec duplication.

index.bs
```diff
@@ -282,30 +288,213 @@ Loader instances are initially created with the internal slots described in the
...
 
 <h2 id="registry">Registry Objects</h2>
 
-<h3 id="registry-abstract-operations">Abstract Operations for Registry Objects</h3>
+<h3 id="registry-constructor">The Registry Constructor</h3>
 
-<h4 id="registry-GetCurrentStage" aoid="GetCurrentStage">GetCurrentStage(entry)</h4>
+The Registry constructor is the %Registry% intrinsic object. When called as a constructor it creates and initializes a new %Registry% object. %Registry% is not intended to be called as a function and will throw an exception when called in that manner.
+
+<h4 id="new-registry" aoid="Registry">Registry()</h4>
```
dherman: Wait -- this function should be the same as `Reflect.prototype.constructor` below. This contradicts the note that says "always throw to prevent the creation of new registry objects in user-land."

index.bs
```diff
@@ -282,30 +288,213 @@ Loader instances are initially created with the internal slots described in the
...
+
+The value of the name property of this function is "[Symbol.iterator]".
+
+<h4 id="registry-prototype-entries">Registry.prototype.entries()</h4>
+
+The following steps are taken:
+
+<emu-alg>
+1. Let _registry_ be *this* value.
+1. If Type(_registry_) is not Object, throw a *TypeError* exception.
```
dherman: All these checks should be checking for the brand (i.e. checking it has all the proper internal properties).

index.bs
dherman: Oh I was just whining. Leave it as-is. :)

index.bs
```diff
@@ -387,233 +606,219 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-1. Let _pair_ be the entry in _entries_ such that _pair_.[[key]] is equal to _key_.
-1. If _pair_ does not exist, then throw a new *TypeError*.
-1. Let _entry_ be _pair_.[[value]].
-1. Let _stageEntry_ be GetCurrentStage(_entry_).
-1. If _stageEntry_.[[Stage]] is "link" or "ready", throw a new *TypeError*.
-1. Remove _pair_ from _entries_.
+1. Let _entry_ be *this* value.
+1. If Type(_entry_) is not Object, return a promise rejected with a new *TypeError* exception.
+1. Let _stageValue_ be ToString(_stage_).
+1. If _stageValue_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
```
dherman: promise rejected with _stageValue_

index.bs
```diff
@@ -387,233 +606,219 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
+1. If IsValidStageValue(_stageValue_) is *false*, return a promise rejected with a new *RangeError* exception.
+1. Let _stageEntry_ be GetStage(_entry_, _stageValue_).
+1. If _stageEntry_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
+1. If _stageEntry_ is *undefined*, return a promise resolved with *undefined*.
+1. If _stageEntry_.[[Result]] is *undefined*, return a promise resolved with *undefined*.
+1. Return the result of transforming _stageEntry_.[[Result]] with a new pass-through promise.
 </emu-alg>
 
-<h4 id="registry-prototype-provide">Registry.prototype.provide(key, stage, value)</h4>
+<h4 id="module-status-prototype-provide">ModuleStatus.prototype.resolve(stage, result)</h4>
```
dherman: id should be "module-status-prototype-resolve"

index.bs
```diff
@@ -387,233 +606,219 @@ Each \[[Stage]] value indicates the currently pending operation. If the \[[Resul
...
-1. Let _entry_ be EnsureRegistered(_loader_, _key_).
-1. Let _stageEntry_ be GetStage(_entry_, _stage_).
-1. If _stageEntry_ is *undefined*, throw a new *TypeError*.
-1. If _stageEntry_.[[Result]] is *undefined*, then set _entry_.[[Result]] to a promise resolved with _value_.
-1. Else, fulfill _entry_.[[Result]] with _value_.
-1. UpgradeToStage(_entry_, _stage_).
+1. Let _entry_ be *this* value.
+1. If Type(_entry_) is not Object, throw a new *TypeError*.
+1. Let _stageValue_ be ToString(_stage_).
+1. If _stageValue_ is an abrupt completion, return a promise rejected with a new *TypeError* exception.
```
dherman: promise rejected with _stageValue_

domenic: Actually _stageValue_.[[value]]. Alternately I think you can just replace this whole step with RejectIfAbrupt(stageValue).

index.bs
caridy: will do

index.bs
caridy: yeah, this is very confusing because we use _stageValue_ as a returned value and as a record of completion. I will use `RejectIfAbrupt(stageValue)`.

index.bs
```diff
@@ -666,21 +889,26 @@ Registry instances are initially created with the internal slots described in th
...
+  1. Set _instance_.[[ModuleStatus]] to _entry_.
   1. For each _dep_ in _instance_.[[RequestedModules]], do:
-    1. Append the record { [[key]]: _dep_, [[value]]: *undefined* } to _deps_.
+    1. Append the record { [[RequestName]]: _dep_, [[Key]]: *undefined*, [[ModuleStatus]]: *undefined* } to _deps_.
 1. Set _entry_.[[Dependencies]] to _deps_.
 1. Set _entry_.[[Module]] to _instance_.
 </emu-alg>
 
+<emu-note>
+  [[Key]] wil identifies the dependency module status in the entry's corresponding loader.
```
dherman: "wil identifies"?

caridy:
Updated HTML: https://rawgit.com/caridy/6c921c618d3cc2b162ef/raw/2528a3224b0962a415fefc65565ac51bac2f2ec0/loader-refactor-registry-entry.html

index.bs
caridy: covered now in step 5 and 6

bmeck:
@caridy still looks like Registry has a Map of ModuleStatus as the inspection mechanism, which is still using Promises like candy for [[Stage]].

caridy:
@bmeck that's correct, but the inspection mechanism is now sync, you can essentially push new modules (resolved module entries) into the registry by creating new `Reflect.Module.Status` objects, and access them via loader, which is always going to be async.

caridy:
fixes https://github.com/whatwg/loader/pull/97

bmeck:
@caridy the registry cannot be inspected synchronously at all if none of the results can be inspected synchronously. it is fine if the user API uses promises for this, but the registry is still a no-go. Other comments around the repo look like they want to make subclassing the loader work with synchronous loading, but this is not possible if all registration results require turning the event loop.

bmeck:
@caridy perhaps you can go through the inspection workflow in less text than the whole spec to show how to synchronously get the results?

caridy:
@bmeck we can have a quick chat, and we can go over the details. just ping me!

index.bs
```diff
@@ -152,117 +152,111 @@ mean the same thing as:
...
+The value of the \[[Prototype]] internal slot of the Loader constructor is the intrinsic object %FunctionPrototype%.
 
-Besides the internal slots and the length property (whose value is 0), the Reflect.Loader constructor has the following properties:
+Besides the internal slots and the length property (whose value is 0), the Loader constructor has the following properties:
 
-<h4 id="Reflect.Loader.prototype">Reflect.Loader.prototype</h4>
+<h4 id="loader-prototype">Loader.prototype</h4>
 
-The value of Reflect.Loader.prototype is an ordinary object.
+The value of Loader.prototype is %LoaderPrototype%.
```
dherman: The initial value of Loader.prototype is the intrinsic object %LoaderPrototype%.

index.bs
```diff
@@ -290,32 +284,250 @@ Loader instances are initially created with the internal slots described in the
...
+<emu-alg>
+1. Let _O_ be ? OrdinaryCreateFromConstructor(Registry, "%RegistryPrototype%", «[[RegistryMap]]» ).
+1. Let _M_ be ObjectCreate(%MapIteratorPrototype%, «[[Map]], [[MapNextIndex]], [[MapIterationKind]]»).
+1. Set _O_’s [[RegistryMap]] internal slot to _M_.
+1. Return _O_.
+</emu-alg>
+
+<h3 id="registry-constructor">The Registry Constructor</h3>
+
+The Registry constructor is the %Registry% intrinsic object. When called as a constructor it creates and initializes a new Registry object. When <b>Registry</b> is called as a function rather than as a constructor, it throws an exception.
```
dherman: This is still inconsistent with Registry.prototype.constructor. As I said before, it should be the case that `Registry === Registry.prototype.constructor`.

index.bs
dherman: Reusing the variable name doesn't really bother me. Lexical scope is lexical scope. :)

dherman:
r+ with one fix for the issue where `Registry !== Registry.prototype.constructor`

PR #119, opened 2015-12-28
[spec] Spec text for ParseExportsDescriptors() and CreateModuleMutator() abtracts

caridy:
### `ParseExportsDescriptors(obj)`

Parses a descriptor object passed into the reflective module record constructor.

Parse as in these examples:

* uninitialized, mutable: { }
* uninitialized, immutable: { const: true }
* initialized, mutable: { value: 42 }
* initialized, immutable: { value: 42, const: true }
* re-export (immutable): { module: m, import: "foo" }

More details here: https://gist.github.com/dherman/fbf3077a2781df74b6d8

### `CreateModuleMutator(module)`

Creates a proxy object to compat-mode with CJS and other module formats.

## End-to-end example:

```js
import * as fs from "fs";

let exports;

// reflective API takes environment descriptors and uses revealing constructor pattern
// to expose mutators to the creator of the module instance object
var mod = new Reflect.Module({
  // initialized mutable (includes "uninitialized" var)
  x: { value: undefined },
  y: { value: 42 },

  // uninitialized mutable (let, class)
  z: { },

  // initialized immutable (const)
  PI: { value: Math.PI, const: true },

  // uninitialized immutable (const)
  TAU: { const: true },

  // re-export
  readFile: { module: fs, import: "readFile" }
}, function(mutator, ns) {
  exports = mutator;
});

// mutator object uses setters to write through to the module's encapsulated environment
console.log(mod.y); // 42
exports.y = 43;
console.log(mod.y); // 43
```

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -684,7 +684,9 @@ The initial value of ModuleStatus.prototype.constructor is the intrinsic object 
 1. Let _entry_ be *this* value.
 1. If Type(_entry_) is not Object, throw a *TypeError* exception.
 1. If _entry_ does not have all of the internal slots of a ModuleStatus Instance (<a href="#module-status-internal-slots">5.5</a>), throw a *TypeError* exception.
-1. Return _entry_.[[Module]].
+1. Let _module_ be _entry_.[[Module]].
+1. If _module_ is a Module Record, return GetModuleNamespace(_module_).
+1. Return *undefined*.
 </emu-alg>
 
 <h4 id="module-status-error">get ModuleStatus.prototype.error</h4>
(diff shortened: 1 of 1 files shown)


matthewrobb:
Is `ModuleMutator` sort of a desugaring for live module bindings?

caridy:
@matthewrobb yes, this PR is all about allowing the creation of module records using an imperative api, to match what you can do with the declarative import and export declarations. Here is an example of what you can do: https://github.com/caridy/loader.js/blob/master/examples/simple.js

matthewp:
So the mutator callback allows you to mutate a module's bindings. Cool. Can you add new properties or only mutate the existing ones?

What is the second callback for, evaluate, I'm guessing it's called to execute the module body?  Any way to modify that on an existing module (assuming it hasn't been evaluated)?  Maybe `mod['[[Evaluate]]'] = function() { ...` ?

caridy:
> So the mutator callback allows you to mutate a module's bindings. Cool. Can you add new properties or only mutate the existing ones?

you can't not add or remove bindings (just like you can't do it with the declarative form), just modify exiting bindings marked as "mutables" in the export descriptor object passed into the constructor of the reflective module record.

> What is the second callback for, evaluate, I'm guessing it's called to execute the module body? Any way to modify that on an existing module (assuming it hasn't been evaluated)? Maybe mod['[[Evaluate]]'] = function() { ... ?

It is an optional callback to match the evaluation step on source text module records. In that callback, you can set new values for some of the bindings in mutator, that's all it is. And yes, it is the internal slot `[[Evaluate]]` on the module record, but that's just internal, users can't modify that in any way.

index.bs
```diff
@@ -684,7 +684,9 @@ The initial value of ModuleStatus.prototype.constructor is the intrinsic object
 1. Let _entry_ be *this* value.
 1. If Type(_entry_) is not Object, throw a *TypeError* exception.
 1. If _entry_ does not have all of the internal slots of a ModuleStatus Instance (<a href="#module-status-internal-slots">5.5</a>), throw a *TypeError* exception.
-1. Return _entry_.[[Module]].
+1. Let _module_ be _entry_.[[Module]].
```
caridy: this was a bug, we should never expose entry.[[Module]], instead, a namespace is what users can consume. with this change, you can inspect any entry, extract their namespace, and use it when creating a reflective module record that re-export from an existing namespace/module.

index.bs
```diff
@@ -1151,26 +1153,73 @@ A <dfn>reflective module record</dfn> is a kind of module record. It extends
...
+
 <h4 id="create-module-mutator" aoid="CreateModuleMutator">CreateModuleMutator(module)</h4>
 
 <emu-alg>
-1. // TODO: spec me
+1. Assert: Assert: _module_ is a Reflective Module Records.
+1. Let _mutator_ be ObjectCreate(%ObjectPrototype%).
+1. Let _env_ be _module_.[[Environment]].
+1. Let _envRec_ be env’s environment record.
+1. For each _name_ in module.[[IndirectExports]], do:
```
guybedford: `_module_.[[IndirectExports]]`?

index.bs
```diff
@@ -1151,26 +1153,73 @@ A <dfn>reflective module record</dfn> is a kind of module record. It extends
...
+  1. Let _g_ be MakeArgGetter(_name_, _envRec_).
+  1. Let _indirectExportDesc_ be the PropertyDescriptor{[[Get]]: _g_, [[Set]]: %ThrowTypeError%, [[Enumerable]]: true, [[Configurable]]: false}.
+  1. Perform ? DefinePropertyOrThrow(_mutator_, _name_, _indirectExportDesc_).
+1. For each _name_ in module.[[LocalExports]], do:
```
guybedford: And again?

guybedford:
I take it these proxy objects wouldn't handle binding circular references? Or is it entirely not possible to create them with the API?

caridy:
@guybedford when creating new dynamic module records, you will have to pass a namespace reference in the export descriptor, which means you need access to it ahead of time, but that namespace can be bound to a source text module record that will fulfill its dependencies later on, so it is entirely posible to have dupes via this API.

Aside from that, remember that re-exports descriptors are immutable, so the proxy only has getter for those.

caridy:
got +1 verbally from @dherman, merging!

index.bs
caridy: will get that fix in master directly since it is just an editorial fix.

PR #120, opened 2016-01-05
Improving Linking Mechanism

caridy:
* `RequestSatisfy` and `RequestLink` resolve to undefined, in preparation for the next refactor.
* `ModuleEvaluation()` have to invoke evaluation once per module, and in all its dependencies.
* If `optionalInstance` is callable, it should return a module namespace exotic object.
* Removing the optional 3rd argument of `ModuleStatus()`, the same can be accomplish by creating a new `ModuleStatus` instance and calling `resolve()` on it.
* `get ModuleStatus.prototype.originalKey` to facilitate inspection of module status objects.
* Accessing the hooks right before using them instead of doing it in a previous turn.
* factories can produce namespace objects that might require declaration and evaluation.
* Editorial fixes and extra validation/assertions.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1145,6 +1145,11 @@ A <dfn>reflective module record</dfn> is a kind of module record. It extends
     <td>The set of re-exported bindings. This ensures that ResolveExport can fully resolve re-exports.</td>
   </tr>
   <tr>
+    <td>\[[RequestedModules]]</td>
+    <td>A List of Module Records.</td>
+    <td>The set of module records requested by re-exported bindings. This ensures that ModuleEvaluation can evaluate each requested module.</td>
+  </tr>
+  <tr>
     <td>\[[Evaluate]]</td>
     <td>A function object or <code>undefined</code></td>
     <td>A thunk to call when the the module is evaluated, or <code>undefined</code> if the module is already evaluated.</td>
@@ -1270,13 +1275,12 @@ Reflective modules are always already instantiated.
 1. Let _func_ be _module_.[[Evaluate]].
 1. Set _module_.[[Evaluated]] to *true*.
 1. Set _module_.[[Evaluate]] to *undefined*.
-1. Assert: _func_ is callable.
-1. For each _dep_ in entry.[[Dependencies]], do:
-  1. Let _depEntry_ be _dep_.[[ModuleStatus]].
-  1. Let _requiredModule_ be _depEntry_.[[Module]].
-  1. Perform ? requiredModule.ModuleEvaluation().
-1. Let _argList_ be a new empty List.
-1. Perform ? Call(func, *undefined*, argList).
+1. For each _requiredModule_ in _module_.[[RequestedModules]], do:
+  1. Assert: _requiredModule_ is a Module Record.
+  1. Perform ? _requiredModule_.ModuleEvaluation().
+1. If IsCallable(_func_) is *true*, then:
+  1. Let _argList_ be a new empty List.
+  1. Perform ? Call(func, *undefined*, argList).
 1. Return *undefined*.
 </emu-alg>
 
@@ -1298,12 +1302,15 @@ When Module is called with arguments <i>descriptors</i>, <i>executor</i>, and <i
 1. Let _localExports_ be a new empty List.
 1. Let _indirectExports_ be a new empty List.
 1. Let _exportNames_ be a new empty List.
+1. Let _requestedModules_ be new empty List.
 1. Let _envRec_ be _env_'s environment record.
 1. For each _desc_ in _exportDescriptors_, do:
   1. Let _exportName_ be _desc_.[[Name]].
   1. Append _exportName_ to _exportNames_.
   1. If _desc_ is an Indirect Export Descriptor, then:
     1. Let _otherMod_ be _desc_.[[Module]].
+    1. If _requestedModules_ does not contain the value of _otherMod_, then:
+      1. Append _otherMod_ to _requestedModules_.
     1. Let _resolution_ be ? _otherMod_.ResolveExport(_desc_.[[Import]], « »).
     1. If _resolution_ is *null*, then throw a *SyntaxError* exception.
     1. Append the record {[[Key]]: _exportName_, [[Value]]: _resolution_} to _indirectExports_.
@@ -1318,10 +1325,7 @@ When Module is called with arguments <i>descriptors</i>, <i>executor</i>, and <i
       1. Call _envRec_.InitializeBinding(_exportName_, _desc_.[[Value]]).
 1. If _evaluate_ is not *undefined*, then
   1. If IsCallable(_evaluate_) is *false*, throw a new *TypeError* exception.
-  1. Let _evaluated_ be *false*.
-1. Else,
-  1. Let _evaluated_ be *true*.
-1. Let _mod_ be a new Reflective Module Record {[[Realm]]: _realm_, [[Environment]]: _env_, [[Namespace]]: *undefined*, [[Evaluated]]: _evaluated_, [[LocalExports]]: _localExports_, [[IndirectExports]]: _indirectExports_, [[Evaluate]]: _evaluate_}.
+1. Let _mod_ be a new Reflective Module Record {[[Realm]]: _realm_, [[Environment]]: _env_, [[Namespace]]: *undefined*, [[LocalExports]]: _localExports_, [[IndirectExports]]: _indirectExports_, [[RequestedModules]]: _requestedModules_, [[Evaluated]]: *false*, [[Evaluate]]: _evaluate_}.
 1. Let _ns_ be ModuleNamespaceCreate(_mod_, _realm_, _exportNames_).
 1. Set _mod_.[[Namespace]] to _ns_.
 1. If _executor_ is not *undefined*, then
--- a/index.bs
+++ b/index.bs
@@ -670,9 +670,9 @@ The initial value of ModuleStatus.prototype.constructor is the intrinsic object 
 1. Return _stageEntry_.[[Stage]].
 </emu-alg>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -601,33 +601,27 @@ The ModuleStatus constructor is the %ModuleStatus% intrinsic object and the init
...
+1. Add new stage entry record { [[Stage]]: "translate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Add new stage entry record { [[Stage]]: "instantiate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Add new stage entry record { [[Stage]]: "satisfy", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Add new stage entry record { [[Stage]]: "link", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Add new stage entry record { [[Stage]]: "ready", [[Result]]: *undefined* } as a new element of the list _pipeline_.
 1. Set _O_’s [[Loader]] internal slot to _loader_.
 1. Set _O_’s [[Pipeline]] internal slot to _pipeline_.
 1. Set _O_’s [[Key]] internal slot to _keyString_.
-1. Set _O_’s [[Module]] internal slot to _module_.
+1. Set _O_’s [[Module]] internal slot to *undefined*.
```
caridy: [[Module]] can't be set during construction anymore, can we done right after view `.resolve()` which is sync.

index.bs
```diff
@@ -676,6 +670,17 @@ The initial value of ModuleStatus.prototype.constructor is the intrinsic object
 1. Return _stageEntry_.[[Stage]].
 </emu-alg>
 
+<h4 id="module-status-module">get ModuleStatus.prototype.key</h4>
```
caridy: adding getter for `key`

index.bs
```diff
@@ -947,8 +950,8 @@ ModuleStatus instances are initially created with the internal slots described i
 1. Let _translateStageEntry_ be GetStage(_entry_, "translate").
 1. If _translateStageEntry_ is *undefined*, return a promise resolved with *undefined*.
 1. If _translateStageEntry_.[[Result]] is not *undefined*, return _translateStageEntry_.[[Result]].
-1. Let _hook_ be GetMethod(_entry_.[[Loader]], @@translate).
 1. Let _p_ be the result of transforming RequestFetch(_entry_) with a fulfillment handler that, when called with argument _payload_, runs the following steps:
+  1. Let _hook_ be GetMethod(_entry_.[[Loader]], @@translate).
```
caridy: accessing the hooks right before using them, not in a previous turn.

index.bs
```diff
@@ -966,8 +969,8 @@ ModuleStatus instances are initially created with the internal slots described i
 1. Let _instantiateStageEntry_ be GetStage(_entry_, "instantiate").
 1. If _instantiateStageEntry_ is *undefined*, return a promise resolved with *undefined*.
 1. If _instantiateStageEntry_.[[Result]] is not *undefined*, return _instantiateStageEntry_.[[Result]].
-1. Let _hook_ be GetMethod(_entry_.[[Loader]], @@instantiate).
 1. Let _p_ be the result of transforming RequestTranslate(_entry_) with a fulfillment handler that, when called with argument _source_, runs the following steps:
+  1. Let _hook_ be GetMethod(_entry_.[[Loader]], @@instantiate).
```
caridy: same as above.

index.bs
```diff
@@ -736,8 +741,7 @@ The following steps are taken:
 1. Let _entry_ be *this* value.
 1. If Type(_entry_) is not Object, return a promise rejected with a new *TypeError* exception.
 1. If _entry_ does not have all of the internal slots of a ModuleStatus Instance (<a href="#module-status-internal-slots">5.5</a>), return a promise rejected with a new *TypeError* exception.
-1. If _stage_ is *undefined* then let _stageValue_ be "fetch".
-1. Else let _stageValue_ be ToString(_stage_).
+1. If _stage_ is *undefined*, let _stageValue_ be "fetch"; otherwise, let _stageValue_ be ToString(_stage_).
```
caridy: one liner for simplicity.

index.bs
```diff
@@ -994,10 +997,8 @@ ModuleStatus instances are initially created with the internal slots described i
       1. Set _pair_.[[Key]] to _depKey_.
       1. Set _pair_.[[ModuleStatus]] to _depEntry_.
       1. Let _currentStageEntry_ be GetCurrentStage(_entry_).
-      1. If _currentStageEntry_.[[Stage]] is "ready", then:
-        1. Return _depEntry_.[[Module]].
-      1. Return the result of transforming RequestSatisfy(_depEntry_) with a fulfillment handler that, when called with value _depEntry_, runs the following steps:
-        1. Return _depEntry_.[[Module]].
+      1. If _currentStageEntry_.[[Stage]] is "ready", return *undefined*.
```
caridy: RequestSatisfy() resolves to *undefined* anyways.

index.bs
```diff
@@ -1080,11 +1080,13 @@ The modules spec should only invoke this operation from methods of Source Text M
 1. Let _deps_ be DependencyGraph(_root_).
 1. For each _dep_ in _deps_, do:
   1. Let _depStageEntry_ be GetCurrentStage(_dep_).
-  1. If _depStageEntry_.[[Stage]] is "link" and _dep_.[[Module]] is a Function object, then:
-    1. Let _f_ be _dep_.[[Module]].
-    1. Let _m_ be ? _f_().
-    1. Set _dep_.[[Module]] to _m_.
-    1. Perform UpgradeToStage(_dep_, "ready").
+  1. If _dep_.[[Module]] is a Function object, then:
```
caridy: correcting steps (from editorial point of view) to evaluate the factory, and taking into consideration that the factory can only return a namespace exotic object.

index.bs
```diff
@@ -1111,7 +1113,7 @@ The modules spec should only invoke this operation from methods of Source Text M
 1. Assert: _entry_ must have all of the internal slots of a ModuleStatus Instance (<a href="#module-status-internal-slots">5.5</a>).
 1. Assert: _result_ must be a List.
 1. If _entry_ is already in _result_, return *undefined*.
-1. Append _entry_ to _result_.
+1. Insert _entry_ as the first element of _result_.
```
caridy: small detail that might be completely irrelevant but it means the deps collection is now in evaluation order.

index.bs
```diff
@@ -1264,9 +1266,18 @@ Reflective modules are always already instantiated.
...
 <emu-alg>
 1. Let _module_ be this Reflective Module Record.
-1. Let _evaluate_ be _module_.[[Evaluate]].
+1. If _module_.[[Evaluated]] is *true*, return *undefined*.
+1. Let _func_ be _module_.[[Evaluate]].
+1. Set _module_.[[Evaluated]] to *true*.
 1. Set _module_.[[Evaluate]] to *undefined*.
-1. Return _evaluate_().
+1. Assert: _func_ is callable.
+1. For each _dep_ in entry.[[Dependencies]], do:
```
caridy: A reflective module might depends on another module that requires evaluation because you can pass a namespace object via re-export descriptor, and we need to guarantee the execution of that piece as well.

caridy: more on this in the next refactor.

index.bs
dherman: Let's call it `ModuleStatus.prototype.originalKey` since it can be deleted from the place it was originally stored in the registry of its original loader. It doesn't need to be a nice short name since it's a rarely used reflective property.

index.bs
```diff
@@ -676,6 +670,17 @@ The initial value of ModuleStatus.prototype.constructor is the intrinsic object
...
 
+<h4 id="module-status-module">get ModuleStatus.prototype.key</h4>
+
+<code>ModuleStatus.prototype.key</code> is an accessor property whose set accessor function is undefined. Its get accessor function performs the following steps:
+
+<emu-alg>
+1. Let _entry_ be *this* value.
+1. If Type(_entry_) is not Object, throw a *TypeError* exception.
+1. If _entry_ does not have all of the internal slots of a ModuleStatus Instance (<a href="#module-status-internal-slots">5.5</a>), throw a *TypeError* exception.
+1. Return _entry_.[[Key]].
```
dherman: We should call the internal slot \[\[OriginalKey\]\] too.

index.bs
dherman: I don't have super strong rationale either way. What you've changed it to is slightly more flexible, since you could choose to change a later hook based on the result of an earlier one. But in practice I don't think you'd want to do that, since it races badly with other concurrently loaded modules. In the end, I doubt it matters much.

index.bs
```diff
@@ -1080,11 +1080,13 @@ The modules spec should only invoke this operation from methods of Source Text M
 1. Let _deps_ be DependencyGraph(_root_).
 1. For each _dep_ in _deps_, do:
   1. Let _depStageEntry_ be GetCurrentStage(_dep_).
-  1. If _depStageEntry_.[[Stage]] is "link" and _dep_.[[Module]] is a Function object, then:
-    1. Let _f_ be _dep_.[[Module]].
-    1. Let _m_ be ? _f_().
-    1. Set _dep_.[[Module]] to _m_.
-    1. Perform UpgradeToStage(_dep_, "ready").
```
dherman: Why doesn't it upgrade to "ready" anymore?

index.bs
dherman: I'm not confident we've nailed the evaluation order semantics yet, but this is fine for now.

dherman:
LGTM, I just have the one question about upgrading, and the one nit of s/key/originalKey/. r+ with those resolved. :shipit:

index.bs
caridy: entries with a factory can produce a module records that requires Declaration Instantiation, which is carry on in the next cycle, where we also upgrade to "ready".

PR #122, opened 2016-01-08
[Refactor] Distinguishing external "stages" from internal "phases"

caridy:
### Features

* "Stages" can be controlled by users via hooks and by calling `.resolve()` or `.reject()` on a ModuleStatus object. Stages are "fetch", "translate" and "instantiate".
* "Phases" (or Evaluation Phases) are internals, and users can't affect or inspect them in any way. Phases are "satisfy", "link" and "ready".
* "Satisfy" phase is triggered before resolving the "instantiate" stage, these is, for now, the last piece that is async in the pipeline, after that, linking and evaluation is sync, but that might change in the future with top level async calls.  
* `Module.evaluate(ns)` is now specced, and it is async to guarantee future proof.
* readiness can only be achieve by invoking `loader.import("name")` or `Module.evaluate(ns)`.

### Other fixes

* update Realm Record term to align with 262.
* mutator getters will throw.

### Invariants

* the namespace of a source text module record can only be accessed in user-land after finishing `SatisfyInstance()` via `await loader.import()`, `await Module.evaluate()` or `get entry.module`.
* the namespace of a dynamic module record is "satisfied" by definition (it can only be created via `new Module()`), which guarantees that all namespace exotic from module records in user-land are "satisfied".
* `new ModuleStatus(loader, key, ns)` can be used to synchronously populate the registry because all namespace objects in user-land are "satisfied".

### Rationale

The primary motivation is to enable synchronous registration of pre-instantiated modules. This is important for use cases like polyfills in the browser and for Node compatibility.

But we would need to be sure that such an API cannot add a not-fully-satisfied module. If we did, there would be no way for the ModuleStatus object in the registry to know where it stands in the loading pipeline. We'd have to keep back-pointers to module namespace objects' ModuleStatus objects in order to know when they are safe to be synchronously added. But there's a many-to-one relationship between ModuleStatus objects and module instances -- and it'd be nonsense to allow different ModuleStatuses to fight with each other about how to satisfy the dependencies of a module.

The way to make this coherent is to have an invariant that by the time a module instance is exposed to userland, all its dependencies have been satisfied. This keeps the roles and responsibilities of module namespace objects and ModuleStatus objects distinct, and ensures that synchronous registration is rational.

### ASCII graph for this pipeline refactor

```
fetch    translate   instantiate
 [ ] ------- [ ] ------- [*]
                        / | \
                       /  |  \
                      /   |   \
                     /    |    \
                    /     |     \
                   /      |      \
                 [*]     [ ]     [ ]
               satisfy   link   ready
```

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -584,9 +584,9 @@ The ModuleStatus constructor is the %ModuleStatus% intrinsic object and the init
 
 The <b>ModuleStatus</b> constructor is designed to be subclassable. It may be used as the value of an <b>extends</b> clause of a class definition. Subclass constructors that intend to inherit the specified <b>ModuleStatus</b> behaviour must include a <b>super</b> call to the <b>ModuleStatus</b> constructor to create and initialize the subclass instance with the corresponding internal slots.
 
-<h4 id="new-module-status" aoid="ModuleStatus">ModuleStatus(loader, key)</h4>
+<h4 id="new-module-status" aoid="ModuleStatus">ModuleStatus(loader, key, ns)</h4>
 
-When ModuleStatus is called with arguments <i>loader</i> and <i>key</i>, the following steps are taken:
+When ModuleStatus is called with arguments <i>loader</i>, <i>key</i> and <i>ns</i>, the following steps are taken:
 
 <emu-alg>
 1. If NewTarget is *undefined*, then throw a *TypeError* exception.
@@ -595,15 +595,26 @@ When ModuleStatus is called with arguments <i>loader</i> and <i>key</i>, the fol
 1. Let _keyString_ be ? ToString(_key_).
 1. Let _O_ be ? OrdinaryCreateFromConstructor(NewTarget, "%ModuleStatusPrototype%", «[[Loader]], [[Pipeline]], [[Key]], [[Module]], [[Metadata]], [[Dependencies]], [[Error]]» ).
 1. Let _pipeline_ be a new List.
-1. Add new stage entry record { [[Stage]]: "fetch", [[Result]]: *undefined* } as a new element of the list _pipeline_.
-1. Add new stage entry record { [[Stage]]: "translate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
-1. Add new stage entry record { [[Stage]]: "instantiate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. If _ns_ is *undefined*, then:
+  1. Let _module_ be *undefined*.
+  1. Let _deps_ be *undefined*.
+  1. Add new stage entry record { [[Stage]]: "fetch", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+  1. Add new stage entry record { [[Stage]]: "translate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+  1. Add new stage entry record { [[Stage]]: "instantiate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Else,
+  1. If _ns_ is not a module namespace exotic object, throw a *TypeError* exception.
+  1. Let _module_ be _ns_.[[Module]].
+  1. Let _deps_ be a new empty List.
+  1. Assert: _module_ is a Module Record.
+  1. Assert: All [[RequestedModule]] of _module_ are safistied.
+  1. Let _result_ be a promise resolved with _ns_.
+  1. Add new stage entry record { [[Stage]]: "instantiate", [[Result]]: _result_ } as a new element of the list _pipeline_.
 1. Set _O_’s [[Loader]] internal slot to _loader_.
 1. Set _O_’s [[Pipeline]] internal slot to _pipeline_.
 1. Set _O_’s [[Key]] internal slot to _keyString_.
-1. Set _O_’s [[Module]] internal slot to *undefined*.
+1. Set _O_’s [[Module]] internal slot to _module_.
 1. Set _O_’s [[Metadata]] internal slot to *undefined*.
-1. Set _O_’s [[Dependencies]] internal slot to *undefined*.
+1. Set _O_’s [[Dependencies]] internal slot to _deps_.
 1. Set _O_’s [[Error]] internal slot to *false*.
 1. Return _O_.
 </emu-alg>
@@ -613,7 +624,7 @@ When ModuleStatus is called with arguments <i>loader</i> and <i>key</i>, the fol
 </emu-note>
 
 <emu-note>
-  The optional second argument module allows to create a module status in a way that synchronously circumvents the asynchronous loading process.
+  The optional third argument ns allows to create a module status in a way that synchronously circumvents the asynchronous loading process.
 </emu-note>
 
 <emu-note>
@@ -759,8 +770,8 @@ The following steps are taken:
 1. Let _p0_ be the result of transforming _result_ with a new pass-through promise.
 1. Let _p1_ be the result of transforming _p0_ with a fulfillment handler that, when called with argument _value_, runs the following steps:
   1. If _stageValue_ is "instantiate", then:
-    1. Perform ? ExtractDependencies(_entry_, _value_, *undefined*).
-    1. Return the result of transforming RequestSatisfy(_entry_, *undefined*) with a fulfillment handler that, when called, runs the following steps:
+    1. Return the result of transforming SatisfyInstance(_entry_, _value_, *undefined*, *undefined*) with a fulfillment handler that, when called with value _instance_, runs the following steps:
+      1. Set _entry_.[[Module]] to _instance_.
       1. Let _stageEntry_ be GetStage(_entry_, _stageValue_).
       1. Assert: _stageEntry_ is not *undefined*.
       1. Fulfill _stageEntry_.[[Result]] with _value_.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -206,7 +206,8 @@ The following steps are taken:
 1. If _loader_ does not have all of the internal slots of a Loader Instance (<a href="#loader-internal-slots">3.5</a>), throw a *TypeError* exception.
 1. Return the result of transforming Resolve(_loader_, _name_, _referrer_) with a fulfillment handler that, when called with argument _key_, runs the following steps:
   1. Let _entry_ be EnsureRegistered(_loader_, _key_).
-  1. Return LoadModule(_entry_, "ready").
+  1. Return the result of transforming LoadModule(_entry_, "instantiate") with a fulfillment handler that, when called, runs the following steps:
+    1. Return EnsureEvaluated(_entry_).
```
caridy: `loader.import()` invoke the completion of the "stages", then invoke the completion of the "phases".

index.bs
```diff
@@ -972,80 +959,38 @@ ModuleStatus instances are initially created with the internal slots described i
...
+    1. Return the result of transforming RequestSatisfy(_entry_, _buckle_) with a fulfillment handler that, when called, runs the following steps:
+      1. Return _optionalInstance_.
 1. Let _pCatch_ be the result of transforming _p_ with a rejection handler that, when called, runs the following steps:
   1. Set _entry_.[[Error]] to *true*.
 1. Set _instantiateStageEntry_.[[Result]] to _p_.
 1. Return _p_.
 </emu-alg>
 
-<h4 id="request-satisfy" aoid="RequestSatisfy">RequestSatisfy(entry)</h4>
+<h4 id="request-satisfy" aoid="RequestSatisfy">RequestSatisfy(entry, buckle)</h4>
```
caridy: `buckle` guarantees that we don't get into circular promises by keeping track of modules pending to finish their instantiation "stage" by reaching the "satisfy" phase.

caridy: we might want to change the name of this abstract operation, because it is not really a "Request" anymore.

index.bs
```diff
@@ -972,80 +959,38 @@ ModuleStatus instances are initially created with the internal slots described i
...
+  1. Append _p_ to _list_.
+1. Return the result of waiting for all _list_.
 </emu-alg>
 
 <emu-note>
-  This abstract operation completes the [[Dependencies]] registry per entry keeping backpointers to each dependency entry and the corresponding keys.
+  This abstract operation guarantees that all [[Dependencies]] are instantiated and ready to be linked and evaluated.
 </emu-note>
 
-<h4 id="request-link" aoid="RequestLink">RequestLink(entry)</h4>
```
caridy: dropped in favor of `EnsureLinked`, and `EnsureEvaluated`

index.bs
```diff
@@ -1338,9 +1292,16 @@ The value of the \[[Prototype]] internal slot of the Module constructor is the i
...
+<h4 id="Module.evaluate">Module.evaluate(ns)</h4>
 
-<b>TODO:</b> way to force evaluation of a module namespace exotic object (<code>Reflect.Module.evaluate(m)</code>? <code>m[Reflect.Module.evaluate]()</code>?)
+<emu-alg>
+1. If _ns_ is not a module namespace exotic object, throw a *TypeError* exception.
+1. Let _module_ be _ns_.[[Module]].
+1. Let _entry_ be _module_.[[Entry]].
+1. Let _status_ be EnsureEvaluated(_entry_).
+1. RejectIfAbrupt(_status_);
+1. Return a promise resolved with *true*.
```
caridy: * probably should resolve to `undefined` because they already have access to the `ns`.
* returning a promise because in the future we might want to hold this until completion of execution (assuming top level away lands).

index.bs
```diff
@@ -773,17 +753,24 @@ The following steps are taken:
 1. Let _stageValue_ be ToString(_stage_).
 1. RejectIfAbrupt(_stageValue_).
 1. If IsValidStageValue(_stageValue_) is *false*, return a promise rejected with a new *RangeError* exception.
+1. Let _stageEntry_ be GetStage(_entry_, _stageValue_).
+1. If _stageEntry_ is *undefined*, return a promise rejected with a new *TypeError* exception.
+1. Perform UpgradeToStage(_entry_, _stageValue_).
```
caridy: If you attempt to `resolve()` or `reject()` a entry, it will be bound to the resolution of the provided value (which is promise based), so we automatically discard any previous "stage".

index.bs
```diff
@@ -773,17 +753,24 @@ The following steps are taken:
...
+      1. Assert: _stageEntry_ is not *undefined*.
+      1. Fulfill _stageEntry_.[[Result]] with _value_.
   1. Else,
+    1. Let _stageEntry_ be GetStage(_entry_, _stageValue_).
+    1. If _stageEntry_ is *undefined*, throw a new *TypeError*.
     1. Fulfill _stageEntry_.[[Result]] with _value_.
-  1. Perform UpgradeToStage(_entry_, _stageValue_).
 1. Let _pCatch_ be the result of transforming _p1_ with a rejection handler that, when called, runs the following steps:
   1. Set _entry_.[[Error]] to *true*.
+1. If _stageEntry_.[[Result]] is *undefined*, set _stageEntry_.[[Result]] to _p1_.
```
caridy: trying to eliminate the racing nature of this by taking over the `[[Result]]` when possible.

index.bs
```diff
@@ -612,18 +595,26 @@ When ModuleStatus is called with arguments <i>loader</i> and <i>key</i>, the fol
...
-1. Add new stage entry record { [[Stage]]: "link", [[Result]]: *undefined* } as a new element of the list _pipeline_.
-1. Add new stage entry record { [[Stage]]: "ready", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. If _ns_ is *undefined*, then:
+  1. Let _module_ be *undefined*.
+  1. Let _deps_ be *undefined*.
+  1. Add new stage entry record { [[Stage]]: "fetch", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+  1. Add new stage entry record { [[Stage]]: "translate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+  1. Add new stage entry record { [[Stage]]: "instantiate", [[Result]]: *undefined* } as a new element of the list _pipeline_.
+1. Else,
+  1. If _ns_ is not a module namespace exotic object, throw a *TypeError* exception.
```
caridy: bringing back the possibility of creating new `ModuleStatus` objects with a module record attached to them in a synchronous way. Inspections on the entry are still async, but this helps nodejs to populate the registry.

index.bs
```diff
@@ -960,7 +956,7 @@ ModuleStatus instances are initially created with the internal slots described i
 1. Return _p_.
 </emu-alg>
 
-<h4 id="request-instantiate" aoid="RequestInstantiate">RequestInstantiate(entry)</h4>
+<h4 id="request-instantiate" aoid="RequestInstantiate">RequestInstantiate(entry, buckle)</h4>
```
dherman: "buckle" isn't a good name. I'd just use something like `instantiateSet`.

index.bs
```diff
@@ -971,81 +967,40 @@ ModuleStatus instances are initially created with the internal slots described i
...
-      1. Return RequestSatisfy(_depEntry_).
-    1. Append _pp_ to _depLoads_.
-  1. Return the result of waiting for all _depLoads_ with a fulfillment handler that, when called, runs the following steps:
-    1. Perform UpgradeToStage(_entry_, "link").
-    1. Return *undefined*.
-1. Let _pCatch_ be the result of transforming _p_ with a rejection handler that, when called, runs the following steps:
-  1. Set _entry_.[[Error]] to *true*.
-1. Set _satisfyStageEntry_.[[Result]] to _p_.
-1. Return _p_.
+1. If _buckle_ is *undefined*, Set _buckle_ is a new empty List.
```
dherman: Set _buckle_ **to** a new empty List.

index.bs
```diff
@@ -891,19 +889,17 @@ ModuleStatus instances are initially created with the internal slots described i
...
 <emu-alg>
 1. Assert: _entry_ must have all of the internal slots of a ModuleStatus Instance (<a href="#module-status-internal-slots">5.5</a>).
-1. Let _instance_ be ? Instantiation(_entry_.[[Loader]], _optionalInstance_, _source_).
 1. Let _deps_ be a new empty List.
 1. If _instance_ is a Module Record, then:
-  1. Assert: _instance_ is a Source Text Module Record.
   1. Set _instance_.[[ModuleStatus]] to _entry_.
-  1. For each _dep_ in _instance_.[[RequestedModules]], do:
-    1. Append the record { [[RequestName]]: _dep_, [[ModuleStatus]]: *undefined* } to _deps_.
+  1. If _instance_ is a Source Text Module Record, then:
```
dherman: Isn't the \[\[RequestedModules\]\] of a dynamic module always an empty list? This conditional seems unnecessary and -- from an OO perspective -- a little gross.

index.bs
dherman: resolve to `undefined` instead of **true**

dherman:
Just had a couple nits I left in line notes. r+ with those addressed. Thanks!!

PR #123, opened 2016-01-09
Status clarifications in README

probins:
There seems to be quite a bit of confusion out there as to the current implementation status (for example recent comments in https://bugs.chromium.org/p/v8/issues/detail?id=1569 ), so here's some suggestions for clarifying how this spec is likely to be implemented.


probins:
I also think this repo should refer to the recent work on whatwg/html for `<script type="module">` - but I'm not sure where: in the spec itself? readme? or just in the roadmap, which is currently the only place where it's mentioned?

PR #127, opened 2016-01-27
mutator should only include setters for writtable or configurable descriptors from [[EnvironmentRecord]]

caridy:


caridy:
closing in favor of PR #129

PR #129, opened 2016-02-21
Export Objects

caridy:
Tagline: Formalizing a reflective API to create export bindings for Reflective Module Records.

## Details

* Removing the confusing `mutator` in favor of re-usable Export Objects
* `Reflect.Module.Export.let()`
* `Reflect.Module.Export.var()`
* `Reflect.Module.Export.const()`
* `Reflect.Module.Export.from()`

## Usage

```javascript
import * as fs from "fs";

var Export = Reflect.Module.Export;

var y, TAU;

// reflective API takes Export objects to represent each export slot
var mod = new Reflect.Module({
  // initialized mutable (includes "uninitialized" var)
  x: Export.var(),
  y: (y = Export.var(42)),

  // uninitialized mutable (let, class)
  z: Export.let(),

  // initialized immutable (const)
  PI: Export.const(Math.PI),

  // uninitialized immutable (const)
  TAU: (TAU = Export.const()),

  // indirect binding from arbitrary object
  bar: Export.from(delegate, "bar"), // optimizable special case; helps interop with existing module systems

  // re-export (heavily optimizable since `fs` is known to be a module namespace object)
  readFile: Export.from(fs, "readFile")
});

console.log(mod.y); // 42
y.set(43);          // mutates the y slot
console.log(mod.y); // 43
TAU.get();          // throws TDZ error
```

## Re-exporing all from Module

```javascript
function exportAllFrom(ns) {
  let o = Object.create(null);
  Object.keys(ns).forEach(key => {
    o[key] = Reflect.Module.Export.from(ns, key);
  });
  return Reflect.Module(o);
} // this return a new namespace object for the new reflective module that was created from the namespace argument.
```

## Node Compat-Mode

```javascript
function exportAllFromCJS(cjs) {
  let o = Object.create(null);
  Object.keys(cjs).forEach(key => {
    o[key] = Reflect.Module.Export.let(cjs[key]);
  });
  return Reflect.Module(o);
} // this return a new namespace object for the new reflective module that was created from the cjs module argument.
```

## Rendered HTML

https://rawgit.com/caridy/6430042924e0436918ea/raw/0f9e2224f1c29ac4449000f5350fdcb2ae439157/loader-pr129.html

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -155,7 +155,7 @@ mean the same thing as:
 1. Return ? OrdinaryDefineOwnProperty(_obj_, _name_, _desc_).
 </emu-alg>
 
-<h3 id="common-operations">Built-in Function Objects</h3>
+<h3 id="sec-built-in-function-objects">Built-in Function Objects</h3>
 
 We follow the ECMA-262 <a href="http://www.ecma-international.org/ecma-262/6.0/index.html#sec-built-in-function-objects">convention for built-in function objects</a> in which the value of NewTarget in each function argument is <b>**undefined**</b> for \[[Call]] and the <i>newTarget</i> parameter for \[[Construct]].
 
@@ -375,8 +375,8 @@ The following steps are taken:
 1. If _registry_ does not have all of the internal slots of a Registry Instance (<a href="#registry-internal-slots">4.4</a>), throw a *TypeError* exception.
 1. Let _M_ be _registry_.[[RegistryMap]].
 1. Let _entries_ be the List that is the value of _M_'s [[MapData]] internal slot.
-1. Repeat for each Record {[[key]], [[value]]} _p_ that is an element of _entries_,
-  1. If _p_.[[key]] is not empty and SameValueZero(_p_.[[key]], _key_) is *true*, return _p_.[[value]].
+1. Repeat for each Record {[[Key]], [[Value]]} _p_ that is an element of _entries_,
+  1. If _p_.[[Key]] is not empty and SameValueZero(_p_.[[Key]], _key_) is *true*, return _p_.[[Value]].
 1. Return *undefined*.
 </emu-alg>
 
@@ -392,11 +392,11 @@ The following steps are taken:
 1. If _entry_ does not have all of the internal slots of a ModuleStatus Instance (<a href="#module-status-internal-slots">5.5</a>), throw a *TypeError* exception.
 1. Let _M_ be _registry_.[[RegistryMap]].
 1. Let _entries_ be the List that is the value of _M_'s [[MapData]] internal slot.
-1. Repeat for each Record {[[key]], [[value]]} _p_ that is an element of _entries_,
-  1. If _p_.[[key]] is not empty and SameValueZero(_p_.[[key]], _key_) is *true*, then
-    1. Set _p_.[[value]] to _entry_.
+1. Repeat for each Record {[[Key]], [[Value]]} _p_ that is an element of _entries_,
+  1. If _p_.[[Key]] is not empty and SameValueZero(_p_.[[Key]], _key_) is *true*, then
+    1. Set _p_.[[Value]] to _entry_.
     1. Return _registry_.
-1. Let _p_ be the Record {[[key]]: _key_, [[value]]: _entry_}.
+1. Let _p_ be the Record {[[Key]]: _key_, [[Value]]: _entry_}.
 1. Append _p_ as the last element of _entries_.
 1. Return _registry_.
 </emu-alg>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1131,7 +1131,7 @@ A <dfn>reflective module record</dfn> is a kind of module record. In addition to
   <tr>
     <td>\[[LocalExports]]</td>
     <td>A List of Strings</td>
-    <td>The set of exported names stored in this module's environment.</td>
+    <td>The set of exported names accessible in this module's environment.</td>
```
caridy: it is not longer stored directly in the env record, but it is accessible from there.

index.bs
```diff
@@ -1145,80 +1145,15 @@ A <dfn>reflective module record</dfn> is a kind of module record. In addition to
   </tr>
 </table>
 
--<h3 id="module-abstract-operations">Abstract Operations for Module Objects</h3>
+<h3 id="module-abstract-operations">Abstract Operations for Module Objects</h3>
 
-<h4 id="parse-exports-descriptors" aoid="ParseExportsDescriptors">ParseExportsDescriptors(obj)</h4>
```
caridy: gone baby gone!

index.bs
```diff
@@ -1229,7 +1164,7 @@ When the abstract operation GetExportNames is called with argument <i>exportStar
 
 <h4 id="resolve-export" aoid="ResolveExport">ResolveExport(exportName, resolveStack, exportStarStack)</h4>
 
-When the abstract operation ResolveExport is called with arguments <i>exportName</i>, <i>resolveStack</i> and <i>exportStarStack</i>, the following steps are taken:
+The ResolveExport concrete method of a Reflective Module Record with arguments <i>exportName</i>, <i>resolveSet</i>, and <i>exportStarSet</i> performs the following steps:
```
caridy: editorial to better align with concrete methods from 262

index.bs
```diff
@@ -1284,45 +1219,36 @@ The Module constructor is the initial value of the Module property of the the Re
 
 The Module constructor is designed to be subclassable. It may be used as the value in an extends clause of a class definition. Subclass constructors that intend to inherit the specified Module behaviour must include a super call to the Module constructor to create and initialize the subclass instance with the internal stage necessary to integrated with loaders.
 
-<h4 id="new-module">Module(descriptors[, executor[, evaluate]])</h4>
+<h4 id="new-module">Module(obj[, evaluate])</h4>
```
caridy: it is not a descriptor anymore, just an object with a bunch of enumerables

index.bs
```diff
@@ -1284,45 +1219,36 @@ The Module constructor is the initial value of the Module property of the the Re
...
+      1. Append the record {[[Key]]: _exportName_, [[Value]]: _resolution_} to _indirectExports_.
     1. Else,
-      1. Assert: _desc_ is a Mutable Export Descriptor.
-      1. Perform ? _envRec_.CreateMutableBinding(_exportName_, *false*).
-    1. If _desc_.[[Initialized]] is *true*, then:
-      1. Call _envRec_.InitializeBinding(_exportName_, _desc_.[[Value]]).
-1. If _evaluate_ is not *undefined*, then
-  1. If IsCallable(_evaluate_) is *false*, throw a new *TypeError* exception.
+      1. Append _exportName_ to _localExports_.
+      1. Call _envRec_.CreateIndirectBinding(_exportName_, _exportObj_.[[target]], _exportObj_.[[targetKey]]).
```
caridy: `CreateIndirectBinding` have to be spec'd in 262, it will be very similar to [CreateImportBinding](https://tc39.github.io/ecma262/#sec-createimportbinding), but it opens the door for Export Objects.

index.bs
```diff
@@ -1284,45 +1219,36 @@ The Module constructor is the initial value of the Module property of the the Re
...
+1. If _evaluate_ was not provided, let _evaluate_ be *undefined*.
+1. Else if IsCallable(_evaluate_) is *false*, throw a new *TypeError* exception.
 1. Let _mod_ be a new Reflective Module Record {[[Realm]]: _realm_, [[Environment]]: _env_, [[Namespace]]: *undefined*, [[LocalExports]]: _localExports_, [[IndirectExports]]: _indirectExports_, [[Evaluated]]: *false*, [[Evaluate]]: _evaluate_, [[HostDefined]]: *undefined*}.
-1. Let _ns_ be ModuleNamespaceCreate(_mod_, _exportNames_).
-1. If _executor_ is not *undefined*, then
```
caridy: mutator is gone now.

index.bs
caridy: here are the necessary changes in 262 to support CreateIndirectBinding:

https://github.com/tc39/ecma262/compare/master...caridy:immutable-indirect-binding?expand=1

rektide:
I'm curious what happened to the descriptors? Why is the API all now exporting objects, no descriptors? It seems like it'd be handy to be handy to be able to getter/setters.

caridy:
@rektide the descriptors we had before were not object property descriptors, they were export descriptors. 

As for getter/setter, we have been drilling on the requirements, eventually we will provide that as a new feature, but that will require spec changes in 262 and discussions with implementers (which we have done some already), which will take longer.

caridy:
We have decided to maintain the old API (export descriptors). Few thoughts:

* it reflects better the construction of a source text module record
* the extra machinery for the special "export module record" is a no-go
* `Export` api has probably better ergonomics, but this is not intended for citizen developers, but low level (libraries and fw) code.
* `Export` doesn't give us the edge we thought in terms of functionalities, everything could be replicated with the export descriptors in the same way you would get it done using declarative syntax.

_note: there are few editorial things from this PR that we will cherry-pick_

caridy:
@dherman did I missed anything else here?

Issue #83, opened 2015-09-24
[task] stage 0: <script type="module">


caridy:

Constellation:
During the implementation, I have encountered the issue, _How "character" / "crossorigin" attributes are handled?_

For example, there is tags that finally loads the same module.

```html
<script type="module" charset="euc-jp" src="A.js"></script>
<script type="module" charset="utf-8" src="B.js"></script>
```

`A.js`
```
import "C.js"
```

`B.js`
```
import "C.js"
```

In the above case, how to apply `charset` and `crossorigin` policies to `C.js`?
Will `C.js` with `euc-jp` be reused in `B.js`'s `import "C.js"`?

One idea I think is that, _inserting the originator(tag)'s attributes to the module key_.
As a result, in the above case, we load `C.js` module twice with the different attributes.

caridy:
jajaja, I don't know, but I'm glad you're raising this issue. Let me circle this back to other folks and see what we can do.

caridy:
In theory (based on the conversation we had today), a and b will be raising (due to the network activity, and A, and B will be also compiled (looking for deps), whichever raises to the completion of the fetch process will issue a fetch for c. This seems to be a real problem, I'm adding `charset` to the notes.

domenic:
We should not allow charset, and always interpret as utf8, like workers do.

Constellation:
How about `crossorigin`? I think it is ok enabling this by default.

caridy:
+1 on that jajaja. If we can settle on that, I can add the corresponding wording into the spec. @domenic, link to that piece in the workers' spec? or any discussion related to this topic in the context of the worker?

Constellation:
And I'm also +1 on utf-8. The module will be rewritten from the old code because it includes non compat syntax and semantics. It's good time to require utf-8.

caridy:
just to clarify my position: my +1 was on the utf8 comment from @domenic

Constellation:
> defer=true by default?

Is it necessary to wait page load completion?

caridy:
right, I should have clarified about the whole question including async/defer. I will add more notes about this tomorrow when I get home. GH mobile is very restrictive.

Constellation:
@caridy:
Awesome!

domenic:
We should get @annevk to help with deciding on `crossorigin`. Maybe we can actually enforce CORS this time.

[...]

PR #134, opened 2016-03-30
Update IE/Edge implementation status

christyharagan:
Small detail, but I noticed the IE/Edge page has updated its status from "Under consideration" to "In development".


teppeis:
Only Edge. IE never supports it.

domenic:
I don't believe they're implementing this spec, but instead script type=module, which is part of the HTML spec.

probins:
that's what the other browsers are currently doing, isn't it? `script type=module` is still listed as a task for this spec (#83) - though ISTM that issue can probably be closed.

PR #136, opened 2016-04-12
Update implementation status for IE

nVitius:
Microsoft has this feature listed as "In development" on their status page.


domenic:
As discussed in https://github.com/whatwg/loader/pull/134, that's not correct.

nVitius:
Noticed that this is a dupe of 134 right after opening. Apologies for that.

PR #137, opened 2016-04-13
Update status of Edge implementation

restlessdesign:
The Edge team has updated this feature’s status to: “In Development”

PR #138, opened 2016-04-13
Update browser vendors status

mischah:
… and link to webkits new and shiny feature status page.

ES6 modules are in development for
* Edge according to https://developer.microsoft.com/en-us/microsoft-edge/platform/status/moduleses6
* Webkit according to https://webkit.org/status/?#feature-modules

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,10 @@ This spec will be implemented in several stages/milestones. See the [roadmap](ht
 
 As the various milestones of the roadmap are completed, browsers will be able to implement them. See the following trackers for the current status of the main browsers:
 
- * IE/Edge: [Under Consideration](https://status.modern.ie/moduleses6?term=modules)
+ * IE/Edge: [In progress](https://status.modern.ie/moduleses6?term=modules)
  * Firefox: [In progress](https://bugzilla.mozilla.org/show_bug.cgi?id=568953)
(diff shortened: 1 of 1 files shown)


mischah:
Sorry, just realized that it’s (partly) an duplicate of #137 and #134 

Would you like me to remove the Edge related changes and do a force push?

domenic:
ES6 modules are not at all equivalent to the loader spec in this repo. See previous discussions in #134, #137, #136.

WebKit does not appear to be working on the loader features in this repo either.

mischah:
Okay. It’s a little confusing 😖

I guess the update to the webkit status page is also obsolete then?
Feel free to close in that case. Otherwise I could update und force push that PR …

matthewp:
Yes, neither refer to the Loader spec. Think about it like this: there is a syntax in JavaScript for importing modules and then there is a mechanism that defines how modules should be loaded.  One is grammar and the other instruction. Those browsers are implementing the former.

matthewrobb:
Does anyone know if the work being done will be exposed to script in a way that could ENABLE loader polyfill? I feel like I know the answer to this but I don't think I quite understand it.

I get that the module hooks leave it up to the environment to provide the "instruction" but if those touch points can be hooked into then we might see this spec evolve more organically.

Namely it seems as though you could spec minimally the bits from this repo that are required to "provide" the module exports.

mischah:
Closing this 😘

PR #144, opened 2016-06-26
Correct links to non-existent section 3.5

probins:


caridy:
hopefully once we move to use ecmakup directly, the autolinking will help.

PR #160, opened 2021-04-30
Add a LICENSE

annevk:
Helps with https://github.com/whatwg/sg/issues/133.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,117 @@
+CC0 1.0 Universal
+
+Statement of Purpose
+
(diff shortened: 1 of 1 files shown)


foolip:
Does it say somewhere in the repo this is the intended license, and this is just clarifying that?

annevk:
No it does not. I'm making it match the LICENSE we used for other documents back then. All contributions but those from @caridy seem minor enough that we don't need explicit consent.

caridy (approved):

